Issue #121, opened 2014-08-19
Backtraces not filtered in Rails


jarrett:
Minitest's default reporter correctly filters backtraces in Rails, but minitest-reporters does not.

I'm using these gem versions:

```
rails 4.1.5
minitest 5.4.0
minitest-rails 2.1.0
minitest-reporters 1.0.5
```

Minitest's default output is:

```
ActionView::Template::Error: can't modify frozen Hash
    app/helpers/markdown_helper.rb:24:in `sanitize_options_for_admin'
    app/helpers/markdown_helper.rb:16:in `markdown_for_admin'
    app/views/resources/show.html.erb:91:in `_app_views_resources_show_html_erb__2321309754475323787_2237192660'
    test/integration/browse_resources_test.rb:40:in `block in <class:BrowseResourcesTest>'
```

With minitest-reporters, the output is:

```
Error:
BrowseResourcesTest#test_freeform_search:
ActionView::Template::Error: can't modify frozen Hash
    /Users/username/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/set.rb:270:in `add'
    /Users/username/rails_app/app/helpers/markdown_helper.rb:24:in `sanitize_options_for_admin'
    /Users/username/rails_app/app/helpers/markdown_helper.rb:16:in `markdown_for_admin'
    /Users/username/rails_app/app/views/resources/show.html.erb:91:in `_app_views_resources_show_html_erb__4055419152259173468_2211015980'
    /Users/username/.rvm/gems/ruby-2.1.0/gems/actionview-4.1.5/lib/action_view/template.rb:145:in `block in render'
    /Users/username/.rvm/gems/ruby-2.1.0/gems/activesupport-4.1.5/lib/active_support/notifications.rb:159:in `block in instrument'
    /Users/username/.rvm/gems/ruby-2.1.0/gems/activesupport-4.1.5/lib/active_support/notifications/instrumenter.rb:20:in `instrument'
    /Users/username/.rvm/gems/ruby-2.1.0/gems/activesupport-4.1.5/lib/active_support/notifications.rb:159:in `instrument'
    /Users/username/.rvm/gems/ruby-2.1.0/gems/actionview-4.1.5/lib/action_view/template.rb:339:in `instrument'
    /Users/username/.rvm/gems/ruby-2.1.0/gems/actionview-4.1.5/lib/action_view/template.rb:143:in `render'
[...]

PR #122, opened 2014-08-29
README addition: Using Rails or custom backtrace filter

jarrett:
This is my proposed README update per #121.


bluerogue251:
Thanks so much!

PR #131, opened 2014-11-11
Add failure summary report to spec_reporter.

tmertens:
@os97673 

Scrolling through hundreds -- or in our case thousands -- of test results to locate the failure message for one or a handful of tests is rather painful.  This adds an error summary report to the end of the spec reporter, but before the final test summary is printed.  It omits skipped tests.

You can view the resulting output of this PR by running `bundle exec rake gallery`, or:

```
  1) Failure: MinitestReportersTest::BadTest#test_a [minitest-reporters/test/gallery/bad_test.rb:6]:
        Expected: 1
          Actual: 2

  2) Failure: MinitestReportersTest::BadTest#test_b [minitest-reporters/test/gallery/bad_test.rb:14]:
        --- expected
        +++ actual
        @@ -1,2 +1,2 @@
         "ab
        -c"
        +d"

  3) Error: MinitestReportersTest::BadTest#test_boom:
        RuntimeError: A random exception
            minitest-reporters/test/gallery/bad_test.rb:18:in `test_boom'

Finished in 0.01229s
6 tests, 5 assertions, 2 failures, 1 errors, 1 skips
```

PR #133, opened 2015-01-08
Allows SpecReporter to only report failures

chrisnicola:
I found the test output a bit noisy in development and it was more useful to just report failures instead of everything.

### DIFF
--- a/lib/minitest/reporters/spec_reporter.rb
+++ b/lib/minitest/reporters/spec_reporter.rb
@@ -52,10 +52,6 @@ module Minitest
         end
       end
 
-      def after_test(test)
-
-      end
-
       protected
 
       def before_suite(suite)
--- a/lib/minitest/reporters/spec_reporter.rb
+++ b/lib/minitest/reporters/spec_reporter.rb
@@ -30,24 +30,40 @@ module Minitest
 
       def record(test)
         super
-        print pad_test(test.name)
-        print_colored_status(test)
-        print(" (%.2fs)" % test.time)
-        puts
-        if !test.skipped? && test.failure
-          print_info(test.failure)
+        if options[:failures_only]
+          if !test.skipped? && test.failure
+            puts test.class.name
+            print pad_test(test.name)
+            print_colored_status(test)
+            print(" (%.2fs)" % test.time)
+            puts
+            print_info(test.failure)
+            puts
+          end
+        else
+          print pad_test(test.name)
+          print_colored_status(test)
+          print(" (%.2fs)" % test.time)
           puts
+          if !test.skipped? && test.failure
+            print_info(test.failure)
+            puts
+          end
         end
+      end
+
+      def after_test(test)
+
       end
 
       protected
 
       def before_suite(suite)
-        puts suite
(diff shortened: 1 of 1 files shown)


chrisnicola:
I should add, the reason printing the test.class.name is needed is that because we only output the name on failures the `puts suite` line in before_suite isn't called normally. So we have to output this for the failed tests.

os97673:
> I should add, the reason printing the test.class.name is needed is that because we only output the name on failures the puts suite line in before_suite isn't called normally. So we have to output this for the failed tests.

As I said it would be better to not print class name every time test is fails, but the only stopper for the PR now is redundant method overriding.  As soon as it will be removed PR will be merged.

chrisnicola:
What would you print instead when the test fails, just the test line? Why wouldn't you print the text context name? It isn't technically a "class name" it is constructed from the "describes" calls. The test name usually helps identify which test file you're calling.

os97673:
I was thinking about caching suite name and printing it upon the first failure.

chrisnicola:
Instead of for each failure from that suite? Yeah that sounds fine. I'll try to do that.

os97673:
> Instead of for each failure from that suite?

yep, exactly.

> Yeah that sounds fine. I'll try to do that.

Thanks for your time.  Let me known when you will either implement this or decide to not do this (in this round)

tmertens:
@chrisnicola Any update on this?  It would be a potential alternative to https://github.com/kern/minitest-reporters/pull/131 for fast test suites.  I'll circle around and try to wrap up that PR today, it got sidelined by a number of other higher priority projects.

My biggest issue with the solution here is the complete lack of feedback when tests are passing.  You could have a test hang in the middle of a long-running test suite and wouldn't know it until well after the tests would otherwise have finished running, since zero output would otherwise indicate tests are running happily in the background.  I wonder if this wouldn't be worth putting as optional output for the Progress formatter instead.

For example, the progress bar is displayed while tests are passing.  When a failure occurs, add one or two blank lines and then print the test info and failure message. Add another blank line or two and begin printing the progress bar again (or the next failure message if multiple tests fail in succession).

Thoughts?  The lack of feedback for passing tests would definitely prohibit me from using this formatter when running our slow, 3000+ test suite locally.

os97673:
Well, I'd prefer to not mix Spec and Progress formatters' functionality.  Though I agree that lack of feedback for passing tests is a valid concern.  Perhaps we should not filter passed tests out but after the suite completed print information about all failed tests?

tmertens:
> I'd prefer to not mix Spec and Progress formatters' functionality.
It could always be a new formatter, subclassed off of Progress (or just as its own class).

> Perhaps we should not filter passed tests out but after the suite completed print information about all failed tests?

That is what is already implemented in #131, I just need to move that formatter to its own file.

os97673:
>> Perhaps we should not filter passed tests out but after the suite completed print information about all failed tests?

> That is what is already implemented in #131, I just need to move that formatter to its own file.

ok, so I'm closing the PR and awaiting a new PR with a new formatter.

PR #135, opened 2015-01-29
Disabe ANSI color on non-TTY terminals

tadman:
When running test suites inside an editor, ANSI code often clutters up the output. Differentiating between terminal and editor execution mode is done with `$stdin.tty?`, a method used in one of the reporters but not shared with the others.

There's a stub in here that replaces ANSI::Code with a pass-through equivalent in non-TTY situations.

This requires minimal changes and seems to make the output a lot cleaner in the editor and still preserves the nice color mode on the terminal.

Relates to Issue#134 and may be a possible solution for that problem as well.

### DIFF
--- a/lib/minitest/reporters.rb
+++ b/lib/minitest/reporters.rb
@@ -7,6 +7,7 @@ module Minitest
   module Reporters
     require "minitest/reporters/version"
 
+    autoload :ANSI, "minitest/reporters/ansi"
     autoload :BaseReporter, "minitest/reporters/base_reporter"
     autoload :DefaultReporter, "minitest/reporters/default_reporter"
     autoload :SpecReporter, "minitest/reporters/spec_reporter"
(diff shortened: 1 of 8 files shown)


os97673:
It looks like there is a global var ($ansi) which control behavior of ANSI's methods the way we want (https://github.com/rubyworks/ansi/blob/master/lib/ansi/code.rb#L3-L7)  IMHO it would be better to use this approach.  WDYT?

tadman:
That could be another solution, though it'd have to be tested.

os97673:
Agree that global var is not an ideal API.  If you believe that you do not like this solution then at least use ANSI::Code#colors (https://github.com/rubyworks/ansi/blob/master/lib/ansi/code.rb#L55-L57) instead of hard-coded list of color methods.

tadman:
Normally I'd agree with you, but the ANSI standard colours haven't changed in decades.

This patch works for me, the tests pass, and I'm sure others might benefit. If you've got a better implementation that would obviously trump this one and I'd be 100% behind it.

os97673:
Ok, I've merged the changes.  Thank you for the contribution.

PR #145, opened 2015-03-30
Fast fail option for SpecReporter

jmarceli:


os97673:
As far as I can see fast_fail for DefaultReporter just stop report on the very first failure, while your implementation halts exist.  And I can not say I like this because to just halt from a normal test execution w/o tearing everything down correctly.
Is it really necessary to halt, perhaps the approach DefaultReporter uses would work for you too?

jmarceli:
Fast fail feature implemented in DefaultReporter is not exactly what I was trying to do, because it is just different formatter for displaying errors. The key point of my modification is just to stop running tests as soon as error appears. Unfortunately I didn't find any other method to achieve this result other than `exit`. I assume that good failing fast implementation shouldn't be the feature of `minitest-reporters` but rather `mintest-rails` gem (it's not about the output format but tests itself). I will try to improve this solution as soon as I get some more free time.

os97673:
Ok, so the resume: reporter should not affect execution of tests, but just provide a nice report for the execution, thus the functionality you want should be implemented in `minitest`, and so I'm closing the PR and the ticket for now.

PR #148, opened 2015-04-29
ProgressReporter: No red if no failures or errors

tjschuck:
Seeing red at the end of a test run is unfortunate when there aren't actually any failures!

Before/after:

This is already how [the SpecReporter behaves](https://github.com/kern/minitest-reporters/blob/6b624d4b75ed07753ca39a87b62a3104a0acfad9/lib/minitest/reporters/spec_reporter.rb#L23-L24).

### DIFF
--- a/lib/minitest/reporters/progress_reporter.rb
+++ b/lib/minitest/reporters/progress_reporter.rb
@@ -66,7 +66,8 @@ module Minitest
         puts
         puts('Finished in %.5fs' % total_time)
         print('%d tests, %d assertions, ' % [count, assertions])
-        print(red { '%d failures, %d errors, ' } % [failures, errors])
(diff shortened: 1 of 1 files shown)


os97673:
just one question: what about skipped?

tjschuck:
I don't have much of an opinion either way.  I suppose I could have also make it green on 0, but I made it yellow to match what already exists in the SpecReporter:



If you'd like, I can make both of them do 0-skipped-as-green as well, but I don't really find the yellow nearly as offensive as the red at the end of a successful run.

Just let me know what you prefer!

bikramwp:
+1

os97673:
Ok, let's keep skipped yellow :)

os97673:

os97673:
new version released

PR #150, opened 2015-05-01
Determines if the terminal supports colors, and if so, enables color output in the reporters.

bikramwp:
When using minitest-reporters in an environment where you can't easily have a TTY, such as using it with the https://github.com/ddollar/foreman gem, falling back to detecting if the terminal supports colors gets colors working again.

### DIFF
--- a/lib/minitest/reporters/ansi.rb
+++ b/lib/minitest/reporters/ansi.rb
@@ -4,7 +4,7 @@ module Minitest
       module Code
 
         def self.color?
-          color_terminal = ENV['TERM'].to_s.downcase.index("color")
+          color_terminal = ENV['TERM'].to_s.downcase.include?("color")
           $stdout.tty? || color_terminal
         end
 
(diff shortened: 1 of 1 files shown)


lib/minitest/reporters/ansi.rb
```diff
@@ -2,7 +2,13 @@ module Minitest
   module Reporters
     module ANSI
       module Code
-        if ($stdout.tty?)
+
+        def self.color?
+          color_terminal = ENV['TERM'].to_s.downcase.index("color")
```
os97673: why `index` instead of `include?` ?

bikramwp: Because it's faster :D   
No reason really, except that index just came to mind.

os97673: please change it to `include?` to make code a little bit more readbable

bikramwp: Sure. Changed.

os97673:
Ok, let's try this ;)

PR #151, opened 2015-05-11
Make minitest's SIGINFO handler work with minitest-reporters

brianstorti:
When `minitest` receives a SIGINFO, it reports the failures that happened so we don't have to wait for all the tests to finish to see which ones broke.
Making `#to_s` call `#report` makes this feature work `minitest-reporters` as well.

That's how it looks when I hit `Ctrl-t` during a test run, before this patch:
screenshot 2015-05-11 14 41 30

That's how it looks after the patch:
screenshot 2015-05-11 14 42 19

### DIFF
--- a/lib/minitest/reporters/base_reporter.rb
+++ b/lib/minitest/reporters/base_reporter.rb
@@ -35,6 +35,10 @@ module Minitest
         after_suite(tests.last.class)
       end
 
+      def to_s
+        report
+      end
+
       protected
 
       def after_suite(test)


os97673:
thank you for the contribution, but I'm not sure that BaseReporter is the right place for the change :(
I've looked at minitest's sources and the only reporter where `#to_s` is overridden is  SummaryReporter (see https://github.com/seattlerb/minitest/blob/master/lib/minitest.rb#L589) and it looks reasonable since not all reporters can report results on every `#to_s` invocation (e.g. check our JUnitReporter).
So, I'd agree that DefaultReporter can provide such functionality while JUnit- or Rubymine- reporters shouldn't.
Thus I'm closing the PR and hope you will provide a new one (and I'm open for discussion if you think I'm wrong)

brianstorti:
@os97673 That's reasonable. Thank you for the feedback. I'll create a new PR adding it just to `DefaultReporter`.

PR #152, opened 2015-05-12
Make minitest's SIGINFO handler work with DefaultReporter

brianstorti:
When minitest receives a `SIGINFO`, it reports the failures that happened so we don't have to wait for all the tests to finish to see which ones broke.
Aliasing `to_s` to `report` makes this feature work with the `DefaultReporter` as well.

Before:
screenshot 2015-05-11 14 41 30

After:
screenshot 2015-05-11 14 42 19


os97673:

PR #155, opened 2015-05-21
Update spec_reporter.rb

type-face:
Added check for test.time.nil? to prevent spec_reporter from breaking when using rails-perftest gem.
 
ERROR: /.rvm/gems/ruby-2.1.4-railsexpress/gems/minitest-reporters-1.0.11/lib/minitest/reporters/spec_reporter.rb:33:in `%': can't convert nil into Float (TypeError)

### DIFF
--- a/lib/minitest/reporters/spec_reporter.rb
+++ b/lib/minitest/reporters/spec_reporter.rb
@@ -30,7 +30,7 @@ module Minitest
         super
         print pad_test(test.name)
         print_colored_status(test)
-        print(" (%.2fs)" % test.time)
+        print(" (%.2fs)" % test.time) unless test.time.nil?
         puts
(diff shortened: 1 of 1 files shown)


os97673:
Perhaps printing 0 would be better than print nothing.  What do you think?

type-face:
Because rails-perftest outputs benchmarking and profiling times, I think indicating a time after the test status might be confusing. See screenshot below. I had also considered removing "PASS" from the output as these are not assertion tests but rather generate performance metrics and reports. I didn't do that, however, as it seemed too specific to rails-perftest. Thoughts?  

screenshot from 2015-05-22 10 01 17

os97673:
I don't see a reason to remove "PASS" from the output.

os97673:

Issue #154, opened 2015-05-21
Distracting SpecReporter output with Shoulda


hakanensari:
With shoulda, SpecReporter outputs:

```
TestTruth
  test_: Truth should be truthy.                  PASS (0.00s)
  test_: Truth should not be falsy.               PASS (0.00s)
``` 

[See here](https://github.com/thoughtbot/shoulda-context/commit/105751d8a281333aaf02eeb2952bb81129ae8847) for why this happens.

Perhaps minitest-reporters should handle this by scrubbing the underscore (or the entire test prefix, which is quite redundant) from the name in `SpecReporter#record`? It's just an extra line and shoulda is fairly popular.

FYI, my current workaround:

```ruby
module Minitest
  module Reporters
    class ShouldaReporter < SpecReporter
      def record(test)
        test.name.gsub!(/^test_: /, "")
        super
      end
    end

    use! ShouldaReporter.new
  end
end
```

os97673:
Well, minitest still has option to run particular test by name and if the reporter will print "modified" name user may have some problems with running the particular one.
From the other hand: I doubt someone will mix both Should and non-Should tests in one repository, so PR is welcome.

PR #156, opened 2015-05-22
Remove underscore in test name output by Shoulda

hakanensari:
See #154.


os97673:

PR #157, opened 2015-05-29
Quit printing exception backtrace twice

gauravmc:
This makes it so that exception backtrace is printed to stdout only once. Looks like this annoying problem has been around for some time. More info inline.

@os97673 @grosser

### DIFF
--- a/lib/minitest/reporters/base_reporter.rb
+++ b/lib/minitest/reporters/base_reporter.rb
@@ -86,8 +86,12 @@ module Minitest
       end
 
       def print_info(e, name=true)
-        print "#{e.exception.class.to_s}: " if name
-        e.message.each_line { |line| print_with_info_padding(line) }
+        if e.is_a?(Minitest::UnexpectedError)
+          message = "#{e.exception.class.to_s}: #{e.exception.message}"
+          message.each_line { |line| print_with_info_padding(line) }
+        else
+          e.message.each_line { |line| print_with_info_padding(line) }
+        end
 
         trace = filter_backtrace(e.backtrace)
         trace.each { |line| print_with_info_padding(line) }


gauravmc:
Actually, this fix is not completely right, changing soon

lib/minitest/reporters/base_reporter.rb
```diff
@@ -85,12 +85,13 @@ def print(*args)
...
-        print "#{e.exception.class.to_s}: " if name
+      def print_info(e, error=true)
         e.message.each_line { |line| print_with_info_padding(line) }
 
-        trace = filter_backtrace(e.backtrace)
-        trace.each { |line| print_with_info_padding(line) }
+        unless error
+          trace = filter_backtrace(e.backtrace)
+          trace.each { |line| print_with_info_padding(line) }
+        end
```
gauravmc: So, the issue is mainly just that `print_info` doesn't need to print backtrace when `e` is an error (which is an instance of `UnexpectedError`), because `UnexpectedError#message` is already much [verbose](https://github.com/seattlerb/minitest/blob/e9752484d624950bb3af4e96fde118d921002dd2/lib/minitest.rb#L699-L702) (contains the original error name, message and backtrace).

It looks like `print_info` is already used (at least by inbuilt reporters) such that the second param is set to `false` when `e` is an assertion, so this should be okay? (hence also renamed that boolean to `error`).

os97673:
It would be very helpful if you'd provide output you have before and after the changes, so it would be obvious what problem you are trying to address.

gauravmc:
@os97673 

For sure.

Before:
```
 test_both_block_settings_and_block_content_tags_must_be_present#ThemeBlockAssetTest (6.95s)
AssetCloud::InvalidBucketError:         AssetCloud::InvalidBucketError: No such namespace: blocks_new/foo.liquid
            /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/buckets/invalid_bucket.rb:17:in `write'
            /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/base.rb:147:in `write'
            /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/callbacks.rb:44:in `write_with_callbacks'
            /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/asset.rb:116:in `store'
            /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/callbacks.rb:44:in `store_with_callbacks'
            /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/validations.rb:21:in `store_with_validation'
            test/unit/asset_cloud/theme_block_asset_test.rb:39:in `block in <class:ThemeBlockAssetTest>'
        /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/buckets/invalid_bucket.rb:17:in `write'
        /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/base.rb:147:in `write'
        /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/callbacks.rb:44:in `write_with_callbacks'
        /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/asset.rb:116:in `store'
        /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/callbacks.rb:44:in `store_with_callbacks'
        /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/validations.rb:21:in `store_with_validation'
        test/unit/asset_cloud/theme_block_asset_test.rb:39:in `block in <class:ThemeBlockAssetTest>'
```


After:

```
 test_both_block_settings_and_block_content_tags_must_be_present#ThemeBlockAssetTest (5.73s)
        AssetCloud::InvalidBucketError: No such namespace: blocks_new/foo.liquid
            /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/buckets/invalid_bucket.rb:17:in `write'
            /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/base.rb:147:in `write'
            /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/callbacks.rb:44:in `write_with_callbacks'
            /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/asset.rb:116:in `store'
            /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/callbacks.rb:44:in `store_with_callbacks'
            /home/vagrant/src/ruby/shopify/vendor/bundle/gems/asset_cloud-2.2.0/lib/asset_cloud/validations.rb:21:in `store_with_validation'
            test/unit/asset_cloud/theme_block_asset_test.rb:39:in `block in <class:ThemeBlockAssetTest>'
```

gauravmc:
@os97673 let me know if you need more info

lib/minitest/reporters/base_reporter.rb
```diff
@@ -85,12 +85,13 @@ def print(*args)
         io.print(*args)
       end
 
-      def print_info(e, name=true)
```
os97673: Why do we need to rename the parameter here?

lib/minitest/reporters/base_reporter.rb
os97673: I'm not sure I like the fix :( perhaps I'm missing something.
Before the fix the method always prints backtraces and also print name if name/error param is true.
After the change it always prints name and prints backtraces only if name/error is true.
Imho it is too much to fix problem with particular exception (and it is unclear why the new code will work with other exceptions).

As for parameter name: I'd prefer to make it describe itself in print_info's context, imho the best name would be `print_name` (or something similar).  `error` parameter of `print_info` says nothing about its meaning for the method (should I pas true of false) and I have to read `print_info` code to understand what it means.  `name` is not the best param, but `error` doesn't improve the situation thus I see no reasons to change it.

gauravmc: > After the change it always prints name and prints backtraces only if name/error is true.
Imho it is too much to fix problem with particular exception (and it is unclear why the new code will work with other exceptions).

That's not right. Minitest wraps all types of exceptions in `UnexpectedError`. See [these lines](https://github.com/seattlerb/minitest/blob/e9752484d624950bb3af4e96fde118d921002dd2/lib/minitest/test.rb#L204-L212). And `UnexpectedError`'s `message` method already contains the backtrace. Hence the change here basically ensures that when an exception happens, this prints name, message and backtrace only once.

`print_name` for that parameter's name sounds good

gauravmc: Ideally you'd want minitest's `Assertion` and `UnexpectedError` classes' `message` to behave the same way, but until then this stops the duplicate output annoyance

gauravmc: @os97673 Have updated the code.. it should be clear now why the change is being made.

lib/minitest/reporters/base_reporter.rb
```diff
@@ -86,8 +86,12 @@ def print(*args)
       end
 
       def print_info(e, name=true)
-        print "#{e.exception.class.to_s}: " if name
-        e.message.each_line { |line| print_with_info_padding(line) }
+        if name
+          print "#{e.exception.class.to_s}: "
+          e.exception.message.each_line { |line| print_with_info_padding(line) }
+        else
+          e.message.each_line { |line| print_with_info_padding(line) }
```
os97673: I'm confused :(  As far as I understand we print e.exception.message to avoid printing backtrace which is in e.message.  Am I right?  If so why name is false we print e.message?  Am I missing something?

gauravmc: `name` is true when `e` is a `UnexpectedError`. `name` is false when `e` is an `Assertion`. You have to look at the definitions of those two classes' `message` method to know why this is the case

os97673: Perhaps it would be better to check for e class explicitly to avoid possible problems if someone will call the method with UnexpectedError and name == false.  Just to make the API easier to use.  What do you  think?

gauravmc: Had thought about it.. little shitty checking for `e`'s class like that, but I'm okay coz `print_info` may as well have been used as you mentioned. Code updated. Lets merge it soon! Seeing exception backtrance printed twice each time is driving me crazy (esp. for long backtraces)..

os97673:
Well, the current version doesn't use name parameter :(

francoisjacques:
@gauravmc I agree with you, there's way too much stuff being reported right now when MiniTest::UnexpectedError is raised.

@os97673 you are also right, name is not being used in @gauravmc last implementation.

Here's a better implementation which satisfies the API requirements and the need to remove the clutter.

```
        print "#{e.exception.class.to_s}: " if name
        e.message.each_line { |line| print_with_info_padding(line) }

        # When e is a Minitest::UnexpectedError, the filtered backtrace is already part of the message printed out
        # by the previous line. In that case, and that case only, skip the backtrace output.
        unless e.is_a?(MiniTest::UnexpectedError)
          trace = filter_backtrace(e.backtrace)
          trace.each { |line| print_with_info_padding(line) }
        end
```

I tried to understand how the reporter got into such a state. By looking at the original version, which we can see in https://github.com/kern/minitest-reporters/commit/3125026c96efe55590bee02cfb14fb35e0550a80. I think perhaps it was made for MiniTest (pre-5), which had a different message?

os97673:
> Here's a better implementation which satisfies the API requirements and the need to remove the clutter.

@francoisjacques your implementation looks ok for me, do you want to create a PR? ;)

> I tried to understand how the reporter got into such a state. By looking at the original version, which we can see in 3125026. I think perhaps it was made for MiniTest (pre-5), which had a different message?

agree, most likely it is pre-5 code which is not optimal now.

francoisjacques:
New pull request opened at https://github.com/kern/minitest-reporters/pull/165

gauravmc:
@francoisjacques just got around to looking at the comments, thanks for taking over

PR #160, opened 2015-06-10
Add ansi color detection for rxvt and xterm

michaelcm:
Here is my attempt at a correction to the ansi color detection for cygwin rxvt and xterm. Thanks!

### DIFF
--- a/lib/minitest/reporters/ansi.rb
+++ b/lib/minitest/reporters/ansi.rb
@@ -4,7 +4,7 @@ module Minitest
       module Code
 
         def self.color?
-          color_terminal = ENV['TERM'].to_s.downcase.include?("color")
+          color_terminal = !!(ENV['TERM'].to_s =~ /(color|rxvt|xterm)/i)
           $stdout.tty? || color_terminal
         end
 
--- a/test/integration/reporters/progress_reporter_test.rb
+++ b/test/integration/reporters/progress_reporter_test.rb
@@ -6,17 +6,17 @@ module MinitestReportersTest
       fixtures_directory = File.expand_path('../../../fixtures', __FILE__)
       test_filename = File.join(fixtures_directory, 'progress_test.rb')
       output = `ruby #{test_filename} 2>&1`
-      assert_match 'ERROR["test_error"', output, 'Errors should be displayed'
-      assert_match 'FAIL["test_failure"', output, 'Failures should be displayed'
-      assert_match 'SKIP["test_skip', output, 'Skipped tests should be displayed'
+      assert_match /ERROR(\033\[0m)*\["test_error"/, output, 'Errors should be displayed'
+      assert_match /FAIL(\033\[0m)*\["test_failure"/, output, 'Failures should be displayed'
+      assert_match /SKIP(\033\[0m)*\["test_skip"/, output, 'Skipped tests should be displayed'
     end
     def test_skipped_tests_are_not_displayed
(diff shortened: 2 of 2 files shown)


os97673:
Travis says that tests are failing after you change :(

michaelcm:
I ran the tests locally and they passed.
```bash
$ ruby -I"lib" -I"C:/ruby/1.9.3/lib/ruby/gems/1.9.1/gems/rake-10.4.2/lib" "C:/ruby/1.9.3/lib/ruby/gems/1.9.1/gems/rake-10.4.2/lib/rake/rake_test_loader.rb" "test/{unit,integration}/**/*_test.rb" 
ansi: 'gem install win32console' to use color on Windows

# Running tests with run options --seed 31415:

..............

Finished tests in 2.090918s, 6.6956 tests/s, 14.8260 assertions/s.


14 tests, 31 assertions, 0 failures, 0 errors, 0 skips
```

I cannot tell from the error what the issue is. To me it looks maybe like it wasn't expecting the color sequences in the output and they are now present because perhaps the terminal the tests are run in supports color by the new definition of color. Please verify and advise how I would account for this in updating the tests?

Thanks!
Michael

michaelcm:
Okay, so two options that seem logical.

1. Have the color detection also look for an environment variable like ENV['TRAVIS_OS_NAME'] that would disable color in automated testing environments.
2. Update spec so that it passes with or without ansi color characters present: 9f8959a32eee1f5419e7951aaecd11cdb3fbce1f

I'm happy to do either, personally the acceptance of color tags in the spec makes more sense to me, but the messages are pretty ugly to read when failures occur.

Opinions?

os97673:
I'd prefer the second option with modifications:
either provide a way for test to control colorization (and disable colors for tests), or "sanitise" output before checking to avoid complicated regexps.  WDYT?

michaelcm:
I'd rather have the test be able to control (disable colors for tests), it seems more direct, though am not sure how to allow the test to communicate with the ansi.rb file or expose out a method or option. The easier to implement is a stripColor method that the test calls on the output it receives.

michaelcm:
Also, if you're willing to make the colorize gem a dependency it has a uncolorize method that does what we need. But seems odd to have a dependency only required to run tests.

os97673:
> But seems odd to have a dependency only required to run tests.

Actually I'm ok with having some additional development requirements.

os97673:
@michaelcm do you plan to work on this any time soon?

michaelcm:
Yes. I've been testing the suggested changes and am not sure they should be pushed in as is. Yes it correctly detects those terminals now, but those terminals don't always support colour, or rather the things those terminals do don't always accept colour. For example, piping results to a file you typically don't want colour characters.

Curious if anyone has any experience in adding an option to mintiest-reporters that can either force enable/disable colour output, and how that option gets passed to minitest-reporters.

Thanks in advance.

os97673:
to do this you need to modify minitest_reporter_pugin.rb (https://github.com/kern/minitest-reporters/blob/master/lib/minitest/minitest_reporter_plugin.rb).
One example on how to add options is https://github.com/rails/rails/blob/master/railties/lib/rails/test_unit/minitest_plugin.rb.  Also there is a good documentation for this in minitest's readme (https://github.com/seattlerb/minitest#writing-extensions)

os97673:
ok @michaelcm , it look like you need more time to work on the PR, so I'm closing it for now, reopen it when it will be ready.
BTW I've recently pushed changes which force non-colored mode for tests.

PR #161, opened 2015-06-12
add an option to JUnitReporter to write the output to a single xml file

andruby:
With this Pull Request, you can initialize the JUnitReporter to write its output to a single xml file instead of one xml file per test suite. This can be useful to simplify or speed up CI parsing.

eg:
```ruby
require "minitest/reporters"
Minitest::Reporters.use! Minitest::Reporters::JUnitReporter.new('reports/path', single_file: true)
```

In the process I also changed the position `empty = true` argument to a keyword argument.

* What do you think would be an appropriate name for the option flag? 
* Is it ok to use keyword arguments? Or should I rewrite it to use a ruby 1.9 style options hash?

### DIFF
--- a/lib/minitest/reporters/junit_reporter.rb
+++ b/lib/minitest/reporters/junit_reporter.rb
@@ -28,7 +28,7 @@ module Minitest
         suites = tests.group_by(&:class)
 
         if @single_file
-          write_xml_file_for("minitest", suites.values.flatten)
+          write_xml_file_for("minitest", tests.group_by(&:class).values.flatten)
         else
           suites.each do |suite, tests|
             write_xml_file_for(suite, tests)
--- a/lib/minitest/reporters/junit_reporter.rb
+++ b/lib/minitest/reporters/junit_reporter.rb
@@ -9,9 +9,10 @@ module Minitest
     # Also inspired by Marc Seeger's attempt at producing a JUnitReporter (see https://github.com/rb2k/minitest-reporters/commit/e13d95b5f884453a9c77f62bc5cba3fa1df30ef5)
     # Also inspired by minitest-ci (see https://github.com/bhenderson/minitest-ci)
     class JUnitReporter < BaseReporter
-      def initialize(reports_dir = "test/reports", empty = true)
+      def initialize(reports_dir = "test/reports", empty = true, options = {})
         super({})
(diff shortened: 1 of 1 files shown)


os97673:
Well, we are supposed to work with at least ruby 1.9, but keyword arguments are ruby 2.0 feature.
Also I'd prefer to be backward compatible, JUnitReporter.new( 'filename', false) still should work.

andruby:
You're right, it's better to keep Ruby 1.9 compatibility. 
I've adapted my code.

lib/minitest/reporters/junit_reporter.rb
```diff
@@ -26,27 +27,36 @@ def report
...
-                        :errors => suite_result[:error_count], :tests => suite_result[:test_count],
-                        :assertions => suite_result[:assertion_count], :time => suite_result[:time]) do
-            tests.each do |test|
-              xml.testcase(:name => test.name, :classname => suite, :assertions => test.assertions,
-                           :time => test.time) do
-                xml << xml_message_for(test) unless test.passed?
-              end
-            end
+        if @single_file
+          write_xml_file_for("minitest", suites.values.flatten)
```
os97673: perhaps it would be better to sort all tests somehow instead of tests.group_by(&:class).values.flattern 
What do you think?

andruby: This way runs the tests in the same order as they would have run before (without single_test). I weary of changing the order the tests are run..

os97673: it doesn't execute the tests but only writes report for them.

andruby: My bad, sorry.

What ordering should we use? If we can use the `group_by(&:class)` that's fine for me. I personally don't care about the ordering as my CI only shows failed tests anyway.

os97673: use class to sort is ok for me too.  The main reason to sort is to get stable order to simplify report comparison.

os97673:
Any change to get final version in near future?

andruby:
@os97673 I've sorted the tests by class.

os97673:
Thank you for the contribution.  I've merged the PR but perhaps you would add some documentation to the class about the new option (so others would know about it ;)

PR #163, opened 2015-06-17
Adding an html reporter

DarynHolmes:
This reporter generates HTML output. 

The report uses bootstrap and JQuery. These files have not been checked in, as they are included in the header section of the report by referencing the CDN sites.

The html file is generated by using ERB. The template is found at templates/index.html.erb.

This works with mintest unit tests and minitest specs. When using minitest specs, the name of the test is slightly modified. It removes the test index so that the report reads slightly better.

The suites\tests are ordered alphabetically and then ordered so that failed test suites\tests are at the top, followed by skipped and then passing tests.   

I am more than willing to make any changes you suggest. 

Thanks, Daryn

### DIFF
--- a/lib/minitest/reporters/html_reporter.rb
+++ b/lib/minitest/reporters/html_reporter.rb
@@ -13,8 +13,8 @@ module Minitest
     #
     # On each test run all files in the reports directory are deleted, this prevents a build up of old reports
     #
-    # The report is generated using using ERB. A custom ERB template can be provided but it is not required
-    # The default ERB template uses JQuery and Boostrap, both of these are included by referencing the CDN sites
+    # The report is generated using ERB. A custom ERB template can be provided but it is not required
+    # The default ERB template uses JQuery and Bootstrap, both of these are included by referencing the CDN sites
     class HtmlReporter < BaseReporter
 
       # The title of the report
@@ -56,10 +56,10 @@ module Minitest
         super({})
 
         defaults = {
-            :title    => 'Test Results',
-            :erb_template    => "#{File.dirname(__FILE__)}/../templates/index.html.erb",
-            :reports_dir => 'test/html_reports',
-            :mode => :safe
+            :title        => 'Test Results',
+            :erb_template => "#{File.dirname(__FILE__)}/../templates/index.html.erb",
+            :reports_dir  => 'test/html_reports',
+            :mode         => :safe
         }
 
         settings = defaults.merge(args)
--- a/lib/minitest/reporters/html_reporter.rb
+++ b/lib/minitest/reporters/html_reporter.rb
@@ -1,0 +1,160 @@
+require 'builder'
+require 'fileutils'
+require 'erb'
+
+module Minitest
+  module Reporters
+    class HtmlReporter < BaseReporter
+
+      attr_reader :title
+
+      def initialize(title = 'Test Results', reports_dir = "test/html_reports", empty = true)
+        super({})
+        @title = title
+        @reports_path = File.absolute_path(reports_dir)
+
+        if empty
+          puts "Emptying #{@reports_path}"
+          FileUtils.remove_dir(@reports_path) if File.exists?(@reports_path)
+          FileUtils.mkdir_p(@reports_path)
+        end
+      end
+
+      def report
+        super
+
+        puts "Writing HTML reports to #{@reports_path}"
+        erb_file = "#{File.dirname(__FILE__)}/../templates/index.html.erb"
+        html_file = @reports_path + "/index.html"
+        erb_str = File.read(erb_file)
+        renderer = ERB.new(erb_str)
+
+        tests_by_suites = tests.group_by(&:class) # taken from the JUnit reporter
+
+        suites = tests_by_suites.map do |suite, tests|
+          suite_summary = summarize_suite(suite, tests)
+          suite_summary[:tests] = tests.sort { |a, b| compare_tests(a, b) }
+          suite_summary
+        end
+
+        suites.sort! { |a, b| compare_suites(a, b) }
+
+        result = renderer.result(binding)
+        File.open(html_file, 'w') do |f|
+          f.write(result)
+        end
+      end
+
+      def passes
+        count - failures - errors - skips
+      end
+
+      def percent_passes
+        100 - percent_skipps - percent_errors_failures # prevetns rounding errors
+      end
+
+      def percent_skipps
+        (skips/count.to_f * 100).to_i
+      end
+
+      def percent_errors_failures
+        ((errors+failures)/count.to_f * 100).to_i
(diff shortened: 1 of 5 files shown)


os97673:
Hi Daryn,

thank you for the contribution.  I like the idea of html presentation, but there are some doubts:

* the reporter has no documentation and it is unclear what result user should expect
* the template is hard-coded, but not everyone will be happy to jquery and bootstrap from CDNs
* please make sure the reporter is used by test:gallery rake task
* I'd consider to use hash instead of three params for initialize e.g. to be able to replace them later with keyword args

This is not the list of "must implement" but "questions to think about".  It will be nice to hear your opinion on them.

DarynHolmes:
Hi and thanks for the feedback.

I can add documentation. Is this the documentation you were referring to: http://www.rubydoc.info/github/CapnKernul/minitest-reporters

If that is where I am meant to add the documentation, then how do I do that? I also have a public blog and I can add a post there with the details if that helps.

The template is hard-coded, it would be nice to have that configurable and allow people to use any template they would like. However I think we can leave that for a future version.

The reason I used the CDN references was to keep the report down to one file. While this does mean the user requires an internet connection, this does makes it easy to email around if needed. Dose this make sense or would be better to include a 'resources' folder which contains Bootstrap and JQuery? 

Another option would be to copy the resources into the report. This way the report would not need an internet connection and would still be only one file. I am not sure if this is a good option. I think the other two are better choices for version one. 

I will make sure that the reporter is used by the test:gallery rake task.

I can make the constructor take a hash. I am going to remove the 'clean' option, and just make it clean the folder on each run. We can always make it keep a history of reports in later versions if required. At the moment, I don't have a need for a history of this report. Jenkins already maintains an adequate history from the JUnit Reporter.

os97673:
> If that is where I am meant to add the documentation, then how do I do that? Is that simply just commenting within the code? I can add comments to the constructor and all the methods, especially the ones that are used by the ERB template, would that help?
> 

I think it is enough to provide documentation for the class (regular code documentation)

> I also have a public blog and I can add a post there with the details and images if that would help.

It is great for initial promotion but I doubt that everyone will read it a year later.

> The template is hard-coded, it would be nice to have that configurable and allow people to use any template they would like. I will have a think about this, it would probably be quite easy to accommodate.
> 
> The reason I used the CDN references was to keep the report down to one file. While this does mean the user requires an internet connection, this does makes it easy to email the report around if needed. Dose this make sense or would it be better to include a 'resources' folder which contains Bootstrap and JQuery?
> 

Current implementation is good for most people and  I think that other problems will be resolved as soon as people will be able to configure the template.

> Another option would be to copy the resources into the reports HTML file. This way the report would not need an internet connection and would still be only one file. I am not sure if this is a good option. I think the other two are better choices for version one.

DarynHolmes:
@os97673 I have made the changes that we discussed. Could you please review it?
Thanks

lib/minitest/reporters/html_reporter.rb
```diff
@@ -0,0 +1,212 @@
...
+    # A reporter for generating HTML test reports
+    # This is recommended to be used with a CI server, where the report is kept as an artifact and is accessible via a shared link
+    #
+    # The reporter sorts the results alphabetically and then by results so that failing and skipped tests are at the top.
+    #
+    # When using Minitest Specs, the number prefix is dropped from the name of the test so that it reads well
+    #
+    # On each test run all files in the reports directory are deleted, this prevents a build up of old reports
+    #
+    # The report is generated using using ERB. A custom ERB template can be provided but it is not required
```
os97673: "using using ERB" ;)

lib/minitest/reporters/html_reporter.rb
```diff
@@ -0,0 +1,212 @@
...
+      # The constructor takes a hash, and uses the following keys:
+      # :title - the title that will be used in the report, defaults to 'Test Results'
+      # :reports_dir - the directory the reports should be written to, defaults to 'test/html_reports'
+      # :erb_template - the path to a custom ERB template, defaults to the supplied ERB template
+      # :mode - Useful for debugging, :terse suppresses errors and is the default, :verbose lets errors bubble up
+      def initialize(args = {})
+        super({})
+
+        defaults = {
+            :title    => 'Test Results',
```
os97673: why do you use several spaces here and on the next line, but just one space on two others?

test/test_helper.rb
```diff
@@ -17,5 +17,5 @@ class TestCase < Minitest::Test
   reporter_klass = Minitest::Reporters.const_get(ENV["REPORTER"])
   Minitest::Reporters.use!(reporter_klass.new)
 else
-  Minitest::Reporters.use!(Minitest::Reporters::DefaultReporter.new)
+  Minitest::Reporters.use! [Minitest::Reporters::DefaultReporter.new, Minitest::Reporters::JUnitReporter.new, Minitest::Reporters::HtmlReporter.new(:title => 'Test Samples')]
```
os97673: Why do you need these changes?

os97673:
Done :)

lib/minitest/reporters/html_reporter.rb
DarynHolmes: Sorry, this is now fixed. I also corrected another typo 'Boostrap'

lib/minitest/reporters/html_reporter.rb
DarynHolmes: This was not deliberate, it has been fixed.

test/test_helper.rb
DarynHolmes: These are not needed, I have removed the changes.

DarynHolmes:
Thanks for the fast feedback, I have committed another version for review. Thanks!

os97673:
Merged, thank you for the contribution.

DarynHolmes:
Great, thanks!

DarynHolmes:
@os97673 I just realised that the readme is not updated. Would you like me to do that?

DarynHolmes:
Is there something I need to do in order to add the documentation here: http://www.rubydoc.info/github/CapnKernul/minitest-reporters

os97673:
> I just realised that the readme is not updated. Would you like me to do that?

yes, please.

os97673:
> Is there something I need to do in order to add the documentation here: http://www.rubydoc.info/github/CapnKernul/minitest-reporters

I do not think so, it is supposed to regenerated automatically (at some point)
Unfortunately I cannot check if the appropriate webhook is configured for the repo (it is not mine :(
@kern could you please clarify how the integration with rubydoc.info is supposed to work for this repo.

kern:
This should be the correct link now: http://www.rubydoc.info/github/kern/minitest-reporters/master

os97673:
@kern  thank you

PR #165, opened 2015-07-06
Quit printing exception backtrace twice, revisited

francoisjacques:
Improves [a fix previously proposed by Gaurav Chande](https://github.com/kern/minitest-reporters/pull/157).

Original comment:
This makes it so that exception backtrace is printed to stdout only
once. Looks like this annoying problem has been around for some time.

### DIFF
--- a/lib/minitest/reporters/base_reporter.rb
+++ b/lib/minitest/reporters/base_reporter.rb
@@ -89,8 +89,12 @@ module Minitest
         print "#{e.exception.class.to_s}: " if name
         e.message.each_line { |line| print_with_info_padding(line) }
 
-        trace = filter_backtrace(e.backtrace)
-        trace.each { |line| print_with_info_padding(line) }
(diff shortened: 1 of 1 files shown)


lib/minitest/reporters/base_reporter.rb
```diff
@@ -89,8 +89,12 @@ def print_info(e, name=true)
         print "#{e.exception.class.to_s}: " if name
```
gauravmc: all types of output from test results is padded.. might wanna use `print_with_info_padding` here too

francoisjacques: @os97673 do you concur?

os97673: Let's keep the change as small as possible.

os97673:
@francoisjacques @gauravmc thank both or you for the contribution.

PR #166, opened 2015-07-06
Disable custom formatters when running under vim

aripollak:
* Vim's quickfix list does not support color codes and only recognizes errors from the
  default Minitest output formatter.


os97673:

Issue #173, opened 2015-09-08
Cannot run tests?


gavinlaking:
Hi,

I've recently forked the project as I have an idea for a reporter I'd like to share with you via a pull request. I've produced it by extracting it from a current project. For my pull request, I'd like to write some tests to prove/show it works, but I am struggling to run the tests.

```
Source/minitest-reporters:feature/mean_time_reporter$ rake gallery
Gemfile Found, running with bundle exec
/home/gavinlaking/.rubies/ruby-2.2.3/bin/ruby: invalid option -2  (-h will show valid options) (RuntimeError)
```

I get the same if I try `rake`, (with or without `bundle exec`), and `rake -2` or `rake -h`. What's the secret sauce?

Kind Regards,

Gav

os97673:
I do nothing special and rake does work for me :(
Perhaps you have some special version of rake?

gavinlaking:
I'm using 10.4.2 by the looks of things. If you don't mind, I'll submit my pull request anyway, and perhaps I can create some tests once you've had a look over it. I'll invoke the "works on my machine" developer card for this! ;-)

Thanks,

Gav

gavinlaking:
My apologies on this issue. I'd derped the `minitest-reporters` gem in my `.gem` directory and `rake` was using the broken version. Hanging my head in shame.

PR #174, opened 2015-09-08
Add 'MeanTimeReporter';

gavinlaking:
Using a temporary 'previous runs' file, we gather statistics on previous
test run times. From this file we build a report which indicates the average
test run time, a minimum (fastest) test run time, a maximum (slowest) test
run time for each test. A description is included for each report line which
is the test name as reported by Minitest.

Tests to follow, pending #173.


os97673:
Thank you for the contribution.  Just a few comments:
- it would be nice to have some description of the reporter (as a class comment)
- could you please add the reporter to the list of reporters which are executed as part of `rake gallery`

lib/minitest/reporters/mean_time_reporter.rb
```diff
@@ -0,0 +1,201 @@
...
+      #   the maximum time the test took to run and a description of the test
+      #   (which is the test name as emitted by Minitest).
+      def report_filename
+        options[:report_filename]
+      end
+
+      # A barbaric way to find out how many runs are in the previous runs file;
+      # this method takes the first test listed, and counts its samples
+      # trusting (naively) all runs to be the same number of samples. This will
+      # produce incorrect averages when new tests are added, so it is advised
```
os97673: Hmm, I'm not sure I like the fact that it is user who is responsible for resetting the data.
Perhaps we could change this and either reset it for her or be able to report stats for both old and new tests.
WDYT?

gavinlaking:
Thanks for the feedback! I'll make the necessary changes.

One consideration which has arisen is how to handle CI. At the moment, TravisCI will not allow you to write to a `/tmp` directory (as there isn't one) to run this reporter, so I'm not sure how to proceed there.

I've also cooked up a `rake` task to reset the statistics file, which reduces some of the burden of resetting the data. I expect that this reporter will be used for many 'runs'. I use this for a project with ~2500 tests and after a couple of hours there are quite a few samples in there. It would be nice to preserve these as a feature could takes days/weeks to build and I might want to see regressions over time. Again, an area that might need a bit of attention.

gavinlaking:
Hi Oleg,

Hope you are well. Do you have any suggestions for how I might demonstrate the functionality of this pull request via tests which are useful to you to prevent regressions and increase the likelihood of a merge?

Gav

os97673:
Actually, with new changes the reporter looks fine for me.  Let's see what users will say ;)

PR #177, opened 2015-09-11
fix NoMethodError

paddor:
lib/minitest/minitest_reporter_plugin.rb:49:in `all_reporters': undefined method `reporters' for Minitest::Reporters:Module (NoMethodError)

### DIFF
--- a/lib/minitest/minitest_reporter_plugin.rb
+++ b/lib/minitest/minitest_reporter_plugin.rb
@@ -1,3 +1,5 @@
+require 'minitest/reporters'
+
 module Minitest
   module Reporters
     class DelegateReporter


os97673:
Could you please describe the situation when the exception is thrown

paddor:
It happens when I run something like `rake test` in various projects, namely https://github.com/mtortonesi/ruby-czmq-ffi and https://github.com/methodmissing/rbczmq. With this change, they don't fail to startup anymore.

paddor:
It happens when I run something like `rake test` in various projects, namely https://github.com/mtortonesi/ruby-czmq-ffi and https://github.com/methodmissing/rbczmq. With this change, they don't fail to startup anymore.

os97673:
Well, I was unable to play with ruby-czmq-ffi, but I see that it can not even run `rake test` with empty gemset, it doesn't specify dependency on minitest and ffi.
As for rbczmq it started w/o problems.  How can I reproduce the problem?

moxley:
@os97673: I got the same error. I was able to resolve it by requiring "minitest/reporters":

```
require "minitest/reporters"
```

os97673:
please provide a test which reproduces the problem.  I'm sure we have the problem I'm not sure the fix is the best one we have.

moxley:
It's not a problem with the gem, it's a problem [between the chair and keyboard](https://en.wiktionary.org/wiki/PEBCAK).

```
gem install minitest-reporters
echo "require 'minitest/autorun'; class MyTest < Minitest::Test; def test_nothing; end; end" > my_test.rb
ruby -I test my_test.rb
```

os97673:
Thank you for the contribution but I've decided to fix this another way.  If minitest reporters are not initialised then the wrapper uses reporters it has got in ctor.  This way we do not affect the behavior for people who has the gem installed but do not use it.

PR #181, opened 2015-09-28
Prevent SpecReporter from changing test name

skuroki:
#156 made SpecReporter#record to change test name which is made by shoulda.

But rails5.x [gets source locations of failed tests with test name](https://github.com/rails/rails/blob/ccee7eea8e5a8a34256467dd1096ac549dd17362/railties/lib/rails/test_unit/reporter.rb#L17).

Then, with rails(edge) + minitest(5.8.0) + SpecReporter(1.1.2) + shoulda(3.5.0), NameError raises when printing failed tests.

This commit will fix it.

### DIFF
--- a/lib/minitest/reporters/spec_reporter.rb
+++ b/lib/minitest/reporters/spec_reporter.rb
@@ -28,8 +28,8 @@ module Minitest
 
       def record(test)
         super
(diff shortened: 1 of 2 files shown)


os97673:

PR #185, opened 2015-10-12
Rounding float values to integers have more value to humans

vadviktor:
I assume a line like this:

`Finished in 0.0007811995047248072h0.04687197028348843m2.812318217009306s, 6.04 tests/s, 10.31 assertions/s`

would have more meaning in this form:

`Finished in 0h0m2s, 6.04 tests/s, 10.31 assertions/s`

so I have rounded those floats down in the report.

### DIFF
--- a/lib/minitest/reporters/html_reporter.rb
+++ b/lib/minitest/reporters/html_reporter.rb
@@ -204,9 +204,9 @@ module Minitest
       def total_time_to_hms
         return ('%.2fs' % total_time) if total_time < 1
 
-        hours = total_time / (60 * 60)
-        minutes = ((total_time / 60) % 60).to_s.rjust(2,'0')
-        seconds = (total_time % 60).to_s.rjust(2,'0')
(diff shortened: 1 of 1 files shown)


os97673:
I agree that it is better to round hours and minutes, but I'm not sure about seconds though.
What perhaps we just need to keep several digits after dot?  What do you think?

vadviktor:
I don't know if anyone would use test reporters as a precise benchmark utility, though I have left the "less than 1 sec" part intact `return ('%.2fs' % total_time) if total_time < 1`

Do you think this has more value when running tests: (note the seconds part)

`Finished in 0h0m2.81s, 6.04 tests/s, 10.31 assertions/s`

than what I suggested above?

IMHO it is meaningless to evaluate e.g. 2.81 over 3.11 when running tests. This is just my idea about testing :)

os97673:
Ok, let's try and see if our ideas work for others ;)

PR #187, opened 2015-10-14
Fix a bug that RE-generated JUnit filenames differently

carols10cents:
Related to #75.

If two describe blocks have the same description, JUnit creates the same
filename for them. To avoid overwriting reports, it adds a number to the
filename if it finds a file already existing.

Before this change, the filename with the number was generated
differently than the original filename-- namely, it did not include the
special character substitution or filename length limiting. Therefore,
if two describes were the same and both used special characters, the
second one would fail because its filename would be invalid.

To fix this, reuse the name generated without the special characters and
limited to a filesystem-friendly length, even when we're adding a number
to avoid filename collisions.

### DIFF
--- a/lib/minitest/reporters/junit_reporter.rb
+++ b/lib/minitest/reporters/junit_reporter.rb
@@ -118,10 +118,11 @@ module Minitest
 
       def filename_for(suite)
         file_counter = 0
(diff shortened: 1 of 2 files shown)


os97673:

PR #188, opened 2015-10-20
Implement DelegateReporter#io

flori:
This makes the duck type of DelegateReporter more similar to
Minitest::StatisticsReporter avoiding a crash if #io is called but not
delegated to one of the reporters. If the io option was passed to
DelegateReporter.new all of the reporters' io attributes should be
initialized to it anyway, so just return it.

### DIFF
--- a/lib/minitest/minitest_reporter_plugin.rb
+++ b/lib/minitest/minitest_reporter_plugin.rb
@@ -7,6 +7,10 @@ module Minitest
         @all_reporters = nil
       end
 
+      def io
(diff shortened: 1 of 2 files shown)


os97673:
Could you please describe the problematic scenario you have?

os97673:
@flori could you please answer my question

os97673:
I believe #189 describes the situation you try to fix.  I've improved the fix a little bit to return io from the first reporter (if we have one)

PR #190, opened 2015-11-02
Substitute `Time.now` for `Minitest.clock_time`;

### LINKED ISSUE #186: Reporters should base time on `Minitest.clock_time`, not `Time.now`
`minitest-reporter`'s (*MR*) `BaseReporter` sometimes reports how far through a test suite a failure happened, overriding `total_time` from `Minitest::StatisticsReporter`.

It does this because Minitest only sets `total_time` at the end of a suite, whereas MR wants to know this earlier (in the case of `fast_fail`, for example).

However, the [overriding method](https://github.com/kern/minitest-reporters/blob/master/lib/minitest/reporters/base_reporter.rb#L68) falls back to `Time.now` as its base to calculate duration, whereas Minitest (since [5.6.1](https://github.com/seattlerb/minitest/blob/master/History.rdoc#561--2015-04-27)) uses a different method, [`Minitest.clock_time`](https://github.com/seattlerb/minitest/blob/78d36807d99404dcbf27bab2091c150ffe3e3f74/lib/minitest.rb#L512), which may have a [different implementation](https://github.com/seattlerb/minitest/blob/78d36807d99404dcbf27bab2091c150ffe3e3f74/lib/minitest.rb#L786) depending on whether `Process::CLOCK_MONOTONIC` is available on the system.

MR should use `Minitest.clock_time` instead of `Time.now` if available, to avoid misreporting times in its reports.

gavinlaking:
Minitest (since 5.6.1) uses `Minitest.clock_time` if `Process::CLOCK_MONOTONIC`
is available on the system. To avoid misreporting times, the implementation has
been changed to use this instead of `Time.now`.

Fixes #186.

### DIFF
--- a/lib/minitest/reporters.rb
+++ b/lib/minitest/reporters.rb
@@ -68,6 +68,18 @@ module Minitest
       end
     end
 
+    def self.clock_time
+      if minitest_version >= 561
+        Minitest.clock_time
+      else
+        Time.now
+      end
+    end
+
+    def self.minitest_version
+      Minitest::VERSION.gsub('.', '').to_i
+    end
+
     def self.use_old_activesupport_fix!
       if defined?(ActiveSupport::VERSION) && ActiveSupport::VERSION::MAJOR < 4
         require "minitest/old_activesupport_fix"
--- a/lib/minitest/reporters/base_reporter.rb
+++ b/lib/minitest/reporters/base_reporter.rb
@@ -66,7 +66,7 @@ module Minitest
       end
 
       def total_time
-        super || Time.now - start_time
+        super || Minitest.clock_time - start_time
       end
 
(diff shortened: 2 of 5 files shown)


os97673:
When Minitest.clock_time is introduced?  Perhaps we should use it only for Minitest >= 5.6.1?

gavinlaking:
Good call, I'll update the pull request to reflect a check on the version of Minitest in use.

gavinlaking:
Sorry for the delay on this.

os97673:
Do not sorry for the work you are doing ;)

gavinlaking:
Just brought this up to date with master. :-)

os97673:

PR #191, opened 2015-11-23
Provide optional sorting for MeanTimeReporter;

gavinlaking:
Provide `:sort_column` and `:order` options;

The `:sort_column` option allows the report summary to be sorted by the Avg, Min, Max or Last column (using `:avg`, `:min`, `:max`, `:last`, respectively). e.g.

`sort_column: :max # => sorts the report summary by the Max column`

The `:order` option allows the report summary to be sorted in either ascending (`:asc` - fastest to slowest) or by default descending order (`:desc` - slowest to fastest).

`order: :asc # => sorts the report summary in ascending order (fastest to slowest)`

### DIFF
--- a/lib/minitest/reporters/mean_time_reporter.rb
+++ b/lib/minitest/reporters/mean_time_reporter.rb
@@ -23,6 +23,9 @@ module Minitest
     #
     class MeanTimeReporter < Minitest::Reporters::DefaultReporter
 
+      class InvalidOrder < StandardError; end
+      class InvalidSortColumn < StandardError; end
+
       # Reset the statistics file for this reporter. Called via a rake task:
       #
       #     rake reset_statistics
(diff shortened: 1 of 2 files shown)


lib/minitest/reporters/mean_time_reporter.rb
```diff
@@ -288,6 +325,58 @@ def rate(run, min, max)
...
+          :asc,
+        ]
+      end
+
+      # @return [Boolean] Whether the given :sort_column option is valid.
+      def valid_sort_column?
+        return true if sort_columns.include?(options[:sort_column])
+
+        fail Minitest::Reporters::MeanTimeReporter::InvalidSortColumn,
+             '`:sort_column` option must be one of `:avg`, `:min`, `:max` or ' \
```
os97673: Imho it would be better to use sort_columns to create error message, so we would need to change only one place to add/remove sort columns

lib/minitest/reporters/mean_time_reporter.rb
```diff
@@ -288,6 +325,58 @@ def rate(run, min, max)
         end
       end
 
+      # @return [Boolean] Whether the given :order option is valid.
+      def valid_order?
+        return true if orders.include?(options[:order])
+
+        fail Minitest::Reporters::MeanTimeReporter::InvalidOrder,
+             '`:order` option must be one of `:desc` or `:asc`'
```
os97673: Though it is unlikely that list of orders will be changed but it would be great to use orders to create error message.

os97673:
@gavinlaking Wow, it looks you've changed something here and I have not received any notification :(
Will review the changes tonight.

os97673:

PR #195, opened 2016-01-23
Improved SpecReporter output

kematzy:
Hi, 

I have made some changes to the SpecReporter which I hope you will agree with and merge into master.

Thanks for your time and attention.

-----

My change retains the default SpecReporter output format as below:

- "test_0001_should return the correct path                        PASS (0.00s)"

But the output can be changed via these configuration options:

* `:show_order`  - sets the output order (:before or :after) [ Default: :after ]
* `:show_time`    - toggle for outputting the timer info.  [ Default: true ]
* `:show_status` - toggle for ouputting the status info   [ Default: true ]

Via a combination of the above alternatives, the following output format can be achieved:

- "should return the correct path"
- "should return the correct path  -  (0.00s)"
- "PASS   should return the correct path"
- "PASS   should return the correct path  -  (0.00s)"

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,48 +56,6 @@ Minitest::Reporters.use! [Minitest::Reporters::DefaultReporter.new(:color => tru
 ```
 
 
-## SpecReporter - configuration options ##
-
-The SpecReporter supports the customised output as highlighted below.
-
-By default the output would look something like this:
-
-```ruby
-"test_0001_should return the correct path                        PASS (0.00s)"
-```
-
-
-But the output can be changed via these configuration options:
-
- *  `:show_order`   - sets the output order (`:before` or `:after`) [ Default: `:after` ]
- 
- *  `:show_time`    - toggle for outputting the timer info.  [ Default: `true` ]
- 
- *  `:show_status`  - toggle for outputting the status info   [ Default: `true` ]
-
-By combining these options, the following output formats can be achieved:
-
-```ruby
-"should return the correct path"
-
-"should return the correct path  -  (0.00s)"
-
-"PASS   should return the correct path"
-
-"PASS   should return the correct path  -  (0.00s)"
-```
-
-Use as follows:
-
-```ruby
-Minitest::Reporters.use! [Minitest::Reporters::SpecReporter.new(:color => true, show_order: :before, ...)]
-
-```
-
-This can aid in the creation of spec documents from your test suite(s).
-
-
-
 ## Screenshots ##
 
 **Default Reporter**
--- a/README.md
+++ b/README.md
@@ -55,6 +55,49 @@ color output from `DefaultReporter`:
 Minitest::Reporters.use! [Minitest::Reporters::DefaultReporter.new(:color => true)]
 ```
 
+
+## SpecReporter - configuration options ##
+
+The SpecReporter supports the customised output as highlighted below.
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -55,6 +55,49 @@ color output from `DefaultReporter`:
...
+By default the output would look something like this:
+
+```ruby
+"test_0001_should return the correct path                        PASS (0.00s)"
+```
+
+
+But the output can be changed via these configuration options:
+
+ *  `:show_order`   - sets the output order (`:before` or `:after`) [ Default: `:after` ]
```
os97673: imho it is unclear what order is controlled here :(
You need either improve the documentation or change option's name.

lib/minitest/reporters/spec_reporter.rb
```diff
@@ -28,10 +38,20 @@ def report
 
       def record(test)
         super
-        test_name = test.name.gsub(/^test_: /, 'test:')
-        print pad_test(test_name)
-        print_colored_status(test)
-        print(" (%.2fs)" % test.time) unless test.time.nil?
+        if options[:show_order] == :before
+           print_colored_status(test) if options[:show_status]
```
os97673: I might be wrong but github shows the changes with a strange indentation :(

lib/minitest/reporters/spec_reporter.rb
```diff
@@ -28,10 +38,20 @@ def report
...
+           test_name = test.name.gsub(/^test_(\d+_)?/, '  ')
```
os97673: I'd expect that test_name will be extracted the same way in both cases.  Why do you use different regexp here?

lib/minitest/reporters/spec_reporter.rb
```diff
@@ -28,10 +38,20 @@ def report
...
+           print(test_name)
+           unless test.time.nil?
+             color = test.time > 0.1 ? :red : :white
+             print(send(color) { "  -  (%.2fs)" } % test.time) if options[:show_time]
+           end
+         else
+          test_name = test.name.gsub(/^test_: /, 'test:')
```
os97673: why :show_time and :show_status used only when :show_order == :before?

os97673:
Please also take a look at Travis' failures.

lib/minitest/reporters/spec_reporter.rb
kematzy: It's indented due to if/else statement.  I'm separating and retaining the current default output in the else statement.  I did not wish to change that part.

lib/minitest/reporters/spec_reporter.rb
kematzy: As above comment.  One version retains your old format, the other uses my new format. 

I'm not clued up or have the time to study possible formats ATM, so I catered "my" version to solving the output I have with my Minitest::Spec based tests. All of which end up with a number in the beginning of them.

os97673: > I'm not clued up or have the time to study possible formats ATM, so I catered "my" version to solving the output I have with my Minitest::Spec based tests. All of which end up with a number in the beginning of them.

than you need to create you own reporter and do not add more options for common one :(

os97673:
the travis build has been fixed.  If the PR is still alive please clean it up.  I'm closing it for now.

PR #196, opened 2016-01-23
Dummy change of version just to test Travis

kematzy:
Hi, 

The whole Travis failures issue seems to have nothing to do with the code I added. No matter what I do it fails with the same error. 

That's why I'm doing this branch just to test Travis.  You can safely ignore this PR.

### DIFF
--- a/lib/minitest/reporters/version.rb
+++ b/lib/minitest/reporters/version.rb
@@ -1,5 +1,5 @@
 module Minitest
   module Reporters
-    VERSION = '1.1.7'
+    VERSION = '1.1.8b1'
   end
 end


kematzy:
As can be seen by this PR and the Travis results produced the errors are still there.

So, just changing the version string from "1.1.7" to  "1.1.8b1"  cannot cause a build error with the following error:

NoMethodError: undefined method `spec' for nil:NilClass
An error occurred while installing minitest-reporters (1.1.8b1), and Bundler cannot continue.
Make sure that `gem install minitest-reporters -v '1.1.8b1'` succeeds before bundling.

Although my Travis foo is almost zero, I think this proves that the problems are on Travis and not in the code I submitted.

It's now 2am and my brain cell needs some off-time, so I leave it at this. Perhaps you can work out what's wrong with Travis.

os97673:
Ok, will try to figure out what could be wrong here :(

PR #198, opened 2016-02-17
JUnit reporter must be careful when removing report dir contents

### LINKED ISSUE #197: Minitest::Reporters::JUnitReporter.new() removes files it shouldn't remove
> rth@rth:~/Software/glowworm$ `rake test`
> Emptying /home/rth/Software/glowworm
> rake aborted!
> Errno::ENOENT: No such file or directory - getcwd
> ...
> rth@rth:~/Software/glowworm$ `ls`
> rth@rth:~/Software/glowworm$ `git status`
> fatal: Unable to read current working directory: No such file or directory

Why not just remove files matching the static pattern in use? Maybe something more like `rm TEST-*-*.xml`?

I think it's quite the assumption that everything in the path provided is fair game for removal.

rthbound:
- No matter what the report directory is specified as,
    the test reporter ought not delete any files it did
    not create itself.
  - Resolves #197 for HtmlReporter and JUnitReporter

### DIFF
--- a/lib/minitest/reporters/html_reporter.rb
+++ b/lib/minitest/reporters/html_reporter.rb
@@ -75,8 +75,8 @@ module Minitest
         @reports_path = File.absolute_path(reports_dir)
 
         puts "Emptying #{@reports_path}"
-        FileUtils.remove_dir(@reports_path) if File.exists?(@reports_path)
         FileUtils.mkdir_p(@reports_path)
+        File.delete(html_file)
       end
 
       # Called by the framework to generate the report
@@ -85,7 +85,6 @@ module Minitest
 
         begin
           puts "Writing HTML reports to #{@reports_path}"
-          html_file = "#{@reports_path}/#{@output_filename}"
           erb_str = File.read(@erb_template)
           renderer = ERB.new(erb_str)
 
@@ -115,6 +114,9 @@ module Minitest
       end
 
       private
+      def html_file
+        "#{@reports_path}/#{@output_filename}"
+      end
 
       def compare_suites_by_name(suite_a, suite_b)
         suite_a[:name] <=> suite_b[:name]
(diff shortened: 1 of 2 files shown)


rthbound:
@os97673 the "empty report directory" strategy this PR aims to fix for the JUnit reporter is in use for other reporters as well.

I'd be glad to make changes to address those instances as well.

rthbound:
Okay, I've made the same change to both: `HtmlReporter` and `JUnitReporter`. It looks like those were the only two reporters to be a little gung-ho on cleaning up after itself.

My thoughts witth this change:

1. only remove files that match the pattern used by the reporter
2. don't remove the directory. if the reporter is responsible for anything, it should only be to make sure the directory exists before trying to write to it.

lib/minitest/reporters/html_reporter.rb
```diff
@@ -75,8 +75,8 @@ def initialize(args = {})
         @reports_path = File.absolute_path(reports_dir)
 
         puts "Emptying #{@reports_path}"
-        FileUtils.remove_dir(@reports_path) if File.exists?(@reports_path)
         FileUtils.mkdir_p(@reports_path)
+        File.delete(*Dir.glob("#{@reports_path}/#{@output_filename}"))
```
os97673: why do you need glob here if this is just one file?

rthbound: Good point. I used the same code as used in the JUnit reporter (which has the possibility of generating more files). I'll amend that.

rthbound:
@os97673 updated...

os97673:
Great.  Just one more request: now we have two places in HtmlReporter where the output file name is constructed.  Could you please extract this code into a separate method to avoid code duplication.

rthbound:
Absolutely :)

rthbound:
How's that, @os97673?

os97673:
Looks fine for me, just give me some time to merge and release a new version.

rthbound:
:+1: Greatly appreciated. Thank you.

PR #199, opened 2016-03-07
Fixing issue with junit reporter attaching wrong test classname

GBH:
it should use classname of the test that ran, not the test runner class.

### DIFF
--- a/lib/minitest/reporters/junit_reporter.rb
+++ b/lib/minitest/reporters/junit_reporter.rb
@@ -48,7 +48,7 @@ module Minitest
                       :errors => suite_result[:error_count], :tests => suite_result[:test_count],
                       :assertions => suite_result[:assertion_count], :time => suite_result[:time]) do
           tests.each do |test|
-            xml.testcase(:name => test.name, :classname => suite, :assertions => test.assertions,
+            xml.testcase(:name => test.name, :classname => test.class, :assertions => test.assertions,
                          :time => test.time) do
               xml << xml_message_for(test) unless test.passed?
             end
--- a/lib/minitest/reporters/progress_reporter.rb
+++ b/lib/minitest/reporters/progress_reporter.rb
@@ -17,7 +17,8 @@ module Minitest
 
       def initialize(options = {})
         super
-        @detailed_skip = options.fetch(:detailed_skip, true)
+        @detailed_skip  = options.fetch(:detailed_skip, true)
+        @slow_treshold  = options.fetch(:slow_treshold , nil)
 
         @progress = ProgressBar.create({
           total:          total_count,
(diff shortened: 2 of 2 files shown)


os97673:
Hmm, I'm not sure why this is a problem :(  The code been the same for very long time (as far as I can see from the version one)  Could you please clarify what the problem is.

GBH:
Sure before my fix you'd get something like this:
```xml
<testcase name="test_redirects_if_wrong_batch_type" classname="minitest" assertions="3" time="0.9684735890004958"></testcase>
<testcase name="test_update" classname="minitest" assertions="4" time="0.11212585900102567"></testcase>
```
It shows what test is being run, but no idea from what file. Everything is `minitest`

Now you can see what file those tests belong to:
```xml
<testcase name="test_get_index" classname="Deals::DealRegistrarsControllerTest" assertions="3" time="0.0300434540004062"></testcase>
<testcase name="test_callback_check_completeness" classname="RegistrationTest" assertions="2" time="0.027196980998269282"></testcase>
```

I was doing integration with CircleCI and noticed completely useless test summary messages. This fixed it.

os97673:
Hmm, I think now I understand the problem and the fix.  It looks ok in the sense it makes the output better.  But imho the real problem is that in one file we have plain structure while it would be better to have node for every suite with sub-nodes  for every test.  Does this look reasonable for you?
If so are you ready (have time and courage) to change the reporter this way?

GBH:
So is this incorrect then? I get that you can structure xml to have a list of testsuites. I guess the question is what should be populated for `name` attribute for `<testsuite>` and `classname` for `<testcase>`?

os97673:
> So is this incorrect then?

Well, since "write to single file" is relatively new feature we still try to find the right way to implement it.  Thus there is no right/wrong.  Your changes makes output better but not perfect (at least I'm not sure the result is perfect, but you can try to convince me that it is ;) )

> I guess the question is what should be populated for name attribute for <testsuite> and classname for <testcase>?

I'd expect that single file would look like
<report>
    <testsuite> ...</testsuite>
      ....
</report>

Where testsuite node is written exactly is it would be written when the reporter creates several files.
Does this make sense for you?  Will this format works for the purpose you are going to use it?
@andruby what you (as author of the functionality) thinks about this?

os97673:
> @os97673 github ate what I assume was an XML snippet. Can you fix that?

done

jules2689:
I just realized this PR was here. I've added this PR https://github.com/kern/minitest-reporters/pull/245 which introduces the same change (more or less) but refactors slightly to adhere to [junit spec format](https://www.ibm.com/support/knowledgecenter/en/SSQ2R2_14.0.0/com.ibm.rsar.analysis.codereview.cobol.doc/topics/cac_useresults_junit.html):

```
<testsuites>
  <testsuite suite="A"></testsuite>
  <testsuite suite="B"></testsuite>
</testsuites>
```

Gwendal35:
Hello,
I use the junit_reporter of v1.1.19 but the commit 90e0420 does not seem to be part of it and so I get the classname="minitest" issue.
            xml.testcase(:name => test.name, :classname => **suite**, :assertions => test.assertions,
                         :time => test.time) do
              xml << xml_message_for(test) unless test.passed?
            end

Is it normal or maybe a regression ?
Thanks in advance

os97673:
> Is it normal or maybe a regression ?

the PR has not been merged yet.

os97673:
Closing in favor of #245

PR #200, opened 2016-03-16
Add HtmlReporter to the readme

TheJefe:
I stumbled onto the html reporter and noticed that it was not included in the readme, so now it is.


os97673:
thank you for the contribution.

PR #205, opened 2016-04-11
Splitting SpecReporter#report for enhanced extensibility

thejamespinto:
Dear Oleg @os97673,

I see you are one of Cucumber's biggest contributors. I think cucumber reports are awesome, but some people/companies don't like to use it for their own reasons. So I built a small gem called "bdd", which brings the benefits of cucumber-style Given/When/Then/And/But reports to non-cucumber users.

I recently started integrating with Minitest but I need to manually reopen SpecReporter so I can add a custom hook in the middle of the name and failure printings so the report will make more sense to the reader (example in the code of this PR)

Here is the output.

image 2016-04-10 at 9 39 04 pm

I don't know if you will like this approach.

Do you think splitting these two printings would help other developers in the future too?

Do you believe there is a better way to accomplish the same result?

Nonetheless, I hope this PR helps other people who want to extend from minitest-reporters,
I would be happy to adapt the code if you have any preferences.

Best,
James.

### DIFF
--- a/lib/minitest/reporters/spec_reporter.rb
+++ b/lib/minitest/reporters/spec_reporter.rb
@@ -28,15 +28,8 @@ module Minitest
 
       def record(test)
         super
-        test_name = test.name.gsub(/^test_: /, 'test:')
-        print pad_test(test_name)
-        print_colored_status(test)
-        print(" (%.2fs)" % test.time) unless test.time.nil?
-        puts
-        if !test.skipped? && test.failure
-          print_info(test.failure)
-          puts
-        end
+        record_print_status(test)
+        record_print_failures_if_any(test)
       end
 
       protected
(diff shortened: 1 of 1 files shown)


os97673:
I'm ok to add the hook but it would be nice to add some documentation to it so everyone would know what can be done in the hook and when it is supposed to be called.
As for better way: could you please provide a link for your reporter otherwise it is hard to suggest anything.

P.S. I am traveling now so please give me some time for response ;)  though I will try to do it as fast as I can

thejamespinto:
Thanks for the quick response!

Here's the link. Branch is still WIP but this file is mostly done.
https://github.com/thejamespinto/bdd/blob/mini/lib/bdd/minitest.rb#L80

thejamespinto:
@os97673 ping 😄

os97673:
sorry for delay (finally back to computer).  I think you second suggestion with `record_print_failures_if_any` and `record_print_status` looks better.

thejamespinto:
@os97673 glad I could be of some help.

PR #209, opened 2016-05-18
Sometimes test.name is a symbol.

craigw:
This means that Builder won't escape it, which allows invalid symbols to sneak into the JUnit output.


os97673:
I might be wrong but looking at Builder's code it looks like it handles symbols correctly (https://github.com/jimweirich/builder/blob/master/lib/builder/xmlmarkup.rb#L323)  Am I missing something?  Perhaps it is a bug in some version of Builder?

os97673:
@craigw I'm closing the PR since there is no response, feel free to reopen it with additional information.

PR #211, opened 2016-07-04
RubyMine reporter provides file URL. Jump to test fixed

avokin:
https://youtrack.jetbrains.com/issue/RUBY-16613  (minitest failure result nodes need "jump to test" functionality). rubymine-reporter provided only name of tested object and testing method's name. It was impossible to restore location of test method. Now reporter provides location.

### DIFF
--- a/lib/minitest/reporters/rubymine_reporter.rb
+++ b/lib/minitest/reporters/rubymine_reporter.rb
@@ -88,8 +88,8 @@ else
 
         def before_test(test)
           super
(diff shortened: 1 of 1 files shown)


os97673:
just to double-check: are you sure your changes will not break integration with previous versions of RubyMine?

avokin:
It shouldn't. locationHint="file:///... supported since early versions of our IDEs

os97673:
released as 1.1.10

PR #212, opened 2016-07-14
Add show_progress option to allow progress output to be turned off

ptrippett:
When using the progress reporter and mean time reporter together, the progress bar is broken by the progress output of the mean time reporter.

This PR adds a `:show_progress` option to the default reporter (defaults to true) allowing the progress to not be displayed.


os97673:
Why you are adding the option do default, imho it should be an option of mean time reporter.
I do not see what other reporter may require such option (if you do please point me out.)

os97673:
I'm closing the PR since no I have not received any response to my request.  Feel free to reopen it with the answer.

PR #213, opened 2016-07-21
MeanTimeReporter: added option to show only running tests

Zhomart:
Otherwise it always shows all tests reports even if I run only 1 test suit.

Configuration
```
# default for :show_all is true, old behavior
Minitest::Reporters::MeanTimeReporter.new(show_all: false)
```

Running one test
```
$ ruby -Ilib:test test/unit/minitest/reporters_test.rb
```


os97673:
IMHO it is not very reliable to filter report text, I'd suggest to change report generation to create the report you need (there you can compare classes with classes)

PR #214, opened 2016-07-22
MeanTimeReporter shows reports only for current running tests

Zhomart:
Otherwise it always shows all tests reports (including reports from other projects) even if I run only 1 test suit.

```
$ ruby -Ilib:test test/unit/minitest/reporters_test.rb
```

### DIFF
--- a/lib/minitest/reporters/mean_time_reporter.rb
+++ b/lib/minitest/reporters/mean_time_reporter.rb
@@ -62,7 +62,7 @@ module Minitest
       def after_suite(suite)
         super
 
-        @all_suite_times = @suite_times
+        @all_suite_times = @suite_times # [ [suite.name:String, time:Float], ..]
       end
 
       # Runs the {Minitest::Reporters::DefaultReporter#report} method and then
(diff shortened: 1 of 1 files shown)


lib/minitest/reporters/mean_time_reporter.rb
```diff
@@ -153,17 +153,21 @@ def order_sorted_body
       # @return [Array<Hash<Symbol => String>>] All of the results sorted by
       #   the :sort_column option. (Defaults to :avg).
       def column_sorted_body
-        previous_run.each_with_object([]) do |(description, timings), obj|
+        current_suite_names = self.tests.each_with_object(Set.new) { |suite, set| set << suite.class.name }
```
os97673: you already have all suite names as keys in current_run hash

lib/minitest/reporters/mean_time_reporter.rb
```diff
@@ -153,17 +153,21 @@ def order_sorted_body
...
           size = Array(timings).size
           sum  = Array(timings).inject { |total, x| total + x }
           obj << {
             avg:  (sum / size).round(9).to_s.ljust(12),
             min:  Array(timings).min.round(9).to_s.ljust(12),
             max:  Array(timings).max.round(9).to_s.ljust(12),
             last: Array(timings).last.round(9).to_s.ljust(12),
-            desc: description,
+            desc: suite_name,
+            suite_name: suite_name
```
os97673: what is the point to add one more field which duplicates existing one?

os97673:
Honestly I do not see the reason to rename description to suire_name all over the code it just make the change bigger but adds nothing.

lib/minitest/reporters/mean_time_reporter.rb
Zhomart: I was thinking description can be changed in the future. It can have shorter test names or can have decorated test names.
Name `suit_name` is for clarity, when we see variable suite_name we exactly know what it is.

Anyway, I can use `description` to filter out suits, and will if you want.

os97673: yes, please, just use description

PR #215, opened 2016-07-24
MeanTimeReporter: added option report_marks, default is true

Zhomart:
By default mean time reporter shows skip/pass/fail marks, which breaks ProgressReporter.
With option `report_marks` now we can disable reporting suite marks.

### DIFF
--- a/lib/minitest/reporters/mean_time_reporter.rb
+++ b/lib/minitest/reporters/mean_time_reporter.rb
@@ -62,7 +62,7 @@ module Minitest
       def after_suite(suite)
         super
 
-        @all_suite_times = @suite_times
+        @all_suite_times = @suite_times # [ [suite.name:String, time:Float], ..]
       end
 
       # Runs the {Minitest::Reporters::DefaultReporter#report} method and then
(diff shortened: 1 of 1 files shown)


Zhomart:
I accidentally added fix for #214 to this PR. Is that ok?

os97673:
I've decided to do this in a little bit different way (see 491b07ddb9527cbd47d23d93e24efda96b9eae52).  Let me know if this works for you and I will release this changes or modify if needed.

Zhomart:
@os97673 Sure. Is it ok just skipping `super` in methods `on_record` and `on_report`?
I'm asking because `BaseReporter` fills out `tests` variable in `BaseReporter#record` method https://github.com/kern/minitest-reporters/blob/master/lib/minitest/reporters/base_reporter.rb#L24.
Also `BaseReporter#report` calls `after_suite(tests.last.class)`.

I think `MeanTimeReporter` should be derived from `BaseReporter`. As `MeanTimeReporter` is deriving from `DefaultReporter`, at least it should call `BaseReporter`'s method everytime, not just skipping them. I could be wrong, because I don't 100% understand how minitest and it's plugins work.

Anyway, do what you think correct, I will create another PR for showing reports only for currently running tests.

os97673:
`on_report`/`on_record`/`on_start` are new methods of `DefaultReporter` I've just added, and they only print some information, so by skipping `supper` in them I just skip this printing, `record`/`report`/`start` of DefaultReporters still call super methods from BaseReporter.

PR #218, opened 2016-10-19
Mean Time Numeric Sort Fixes

matt-glover:
Background
------------
While investigating some extremely slow tests using the `MeanTimeReporter` I spotted some of the slowest results in the middle of the report. I investigated and discovered that the reporter sorts the output against formatted string columns rather than the underlying numbers.

Change
--------
This pull sorts against the numeric values before converting them to strings for the final report.

Other Notes
------------
I am definitely open to alternative ways to test this. I came up with something that seems to work for now without a refactor of the reporter.


gavinlaking:
LGTM :+1:

PR #224, opened 2016-12-15
NUT-3192 port nutrio reporter to our gem

midwire:
Ported our old nutrio minitest reporter to the gem and modified the report line with better time outputs according to specs in the NUT-3192 ticket. The gem itself does not test the custom reporters and I did not add tests for ours either but I've manually tested it with our test runs.


midwire:
Sorry, did not mean to post this to the `kern` repo.

PR #227, opened 2017-01-11
Move HtmlReporter deletion to #start

icantrap:
- Moving the deletion of the HTML report from constructor to
  #start hook ensures that any existing HTML report doesn't
  get wiped when being used with RubyMine reporter.


os97673:

Issue #168, opened 2015-07-23
rails5


elfassy:
I can't get it to work with the master branch of rails (rails 5.0.0.alpha). There is just no color.
Anyone else have this problem?

os97673:
I (almost) do not use rails and thus I have not seen these problems.  But I'm ready to fix them as soon as you will provide a simple repo to reproduce the problem.

elfassy:
Here you go: https://github.com/elfassy/minitest_reporter_rails5

os97673:
I've just cloned the repo, "bundle install", "rake test" and everything works ok :(
What have I missed?

elfassy:
did you get color and ProgressReporter to work?

os97673:
Oops, I've expected exception (thought that this is another ticket :(
No I see the problem.  The real problem is that minitest-reporters are not used at all.

os97673:
Filed rails/rails#21102.  Will see response of rails developers.

os97673:
rails/rails#21105 has been merged thus the problem should be fixed

elfassy:
Thanks @os97673 
Unfortunately that rails fix doesn't see to do it for me. I've updated the example code. It's still giving me the same result (reporters are not used). Did you get the test app to work on your end?

os97673:
You are right, it looks like I need to do some other changes too (I had modified version of the test project and minitest-reporters)

os97673:
I've created #171 with possible fix for the problem.  Could you please try one and let me know if it works for you?

jonatack:
Thanks @os97673 for looking at this. FWIW I just tried it using #171 with Rails 5 (`gem 'minitest-reporters', github: 'kern/minitest-reporters', branch: 'rails5-support'`) and the reporter is not being used.

os97673:
@jonatack please provide the test project which demonstrates the problem.  The test provided by @elfassy with the branch works fine for me. (btw make sure you have up-to-date rails5 and you have stopped spring before the testing)

PR #231, opened 2017-08-07
Ensure DelegateReporter behaves on SIGINFO

joshpencheon:
This issue was addressed before, in PR #152.

However, since then, the `DelegateReporter` wrapper class was added in 25d19942 in response to issue #168.

This PR adds in a similar fix to #152 to ensure the desired behaviour continues to work.

### DIFF
--- a/lib/minitest/minitest_reporter_plugin.rb
+++ b/lib/minitest/minitest_reporter_plugin.rb
@@ -26,6 +26,8 @@ module Minitest
         all_reporters.each(&:report)
       end
 
+      alias to_s report
+
       def passed?
         all_reporters.all?(&:passed?)
       end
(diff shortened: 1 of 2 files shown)


os97673:
Could you please add a test to make sure we will not break it in future.

joshpencheon:
Dear Oleg,

Happy to, but do you have any pointers? I think that SIGINFO isn't supported on Linux (and therefore Travis), so I'm not sure how best to proceed?

I can add a direct test for #to_s, but thought an integration test would be good too if feasible...

os97673:
well, why to_s is important here? Perhaps it is enough to check that report has to_s correctly implemented?

joshpencheon:
@os97673 - I've pushed a test case. Do you feel it provides adequate coverage?

os97673:
Looks reasonable.  Thank you for the contribution.

os97673:
Have to revert you changes because of #233 :(

PR #232, opened 2017-08-11
Fix progress bar when using filter and specs

adaedra:
* Make the correct calculation when filtering on minitest/spec and using the progress reporter

When running with `minitest/spec` style tests and using filtering on class part, the tests are correctly executed, but the progress bar reporter doesn't. The total count of tests is at 0, and the output is the following : 

```
  0/0: [===================================] 100% Time: 00:00:00, Time: 00:00:00
WARNING: Your progress bar is currently at 0 out of 0 and cannot be incremented. In v2.0.0 this will become a ProgressBar::InvalidProgressError.
  0/0: [===================================] 100% Time: 00:00:01, Time: 00:00:01
```

This patch ensures the right count is done and as such the output is correct.

### DIFF
--- a/lib/minitest/minitest_reporter_plugin.rb
+++ b/lib/minitest/minitest_reporter_plugin.rb
@@ -36,9 +36,9 @@ module Minitest
         filter = options[:filter] || '/./'
         filter = Regexp.new $1 if filter =~ /\/(.*)\//
 
-        Minitest::Runnable.runnables.map(&:runnable_methods).flatten.find_all { |m|
-          filter === m || filter === "#{self}##{m}"
-        }.size
+        Minitest::Runnable.runnables.map { |runnable|
+          runnable.runnable_methods.map { |m| "#{runnable}##{m}" }
+        }.flatten.grep(filter).size
       end
 
       def all_reporters
--- a/lib/minitest/minitest_reporter_plugin.rb
+++ b/lib/minitest/minitest_reporter_plugin.rb
@@ -37,8 +37,10 @@ module Minitest
         filter = Regexp.new $1 if filter =~ /\/(.*)\//
 
         Minitest::Runnable.runnables.map { |runnable|
-          runnable.runnable_methods.map { |m| "#{runnable}##{m}" }
-        }.flatten.grep(filter).size
+          runnable.runnable_methods.find_all { |m|
+            filter === m || filter === "#{runnable}##{m}"
+          }.size
+        }.inject(:+)
       end
 
       def all_reporters
--- a/test/fixtures/spec_test.rb
+++ b/test/fixtures/spec_test.rb
@@ -1,0 +1,13 @@
+require 'minitest'
+require 'minitest/reporters'
+require 'minitest/autorun'
+
+Minitest::Reporters.use! Minitest::Reporters::ProgressReporter.new
+
+describe String do
+  describe '#length' do
+    it 'works' do
+      assert_equal 5, 'hello'.length
+    end
+  end
+end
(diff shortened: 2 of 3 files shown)


adaedra:
Hello, @kern @os97673 ?

lib/minitest/minitest_reporter_plugin.rb
```diff
@@ -34,11 +34,11 @@ def passed?
...
         filter = options[:filter] || '/./'
-        filter = Regexp.new $1 if filter =~ /\/(.*)\//
+        filter = Regexp.new $1 if filter =~ %r"/(.*)/"
 
-        Minitest::Runnable.runnables.map(&:runnable_methods).flatten.find_all { |m|
-          filter === m || filter === "#{self}##{m}"
-        }.size
+        Minitest::Runnable.runnables.map do |runnable|
+          runnable.runnable_methods.map { |meth| "#{runnable}##{meth}" }
+        end.flatten.grep(filter).size
```
os97673: could you please avoid stylistic changes

adaedra:
Hi @os97673, I applied the changes, is it ok?

lib/minitest/minitest_reporter_plugin.rb
```diff
@@ -36,9 +36,9 @@ def total_count(options)
         filter = options[:filter] || '/./'
         filter = Regexp.new $1 if filter =~ /\/(.*)\//
 
-        Minitest::Runnable.runnables.map(&:runnable_methods).flatten.find_all { |m|
-          filter === m || filter === "#{self}##{m}"
-        }.size
+        Minitest::Runnable.runnables.map { |runnable|
+          runnable.runnable_methods.map { |m| "#{runnable}##{m}" }
```
os97673: Am I right that the important change is replacing "#{self}##{m}" with "#{runnable}##{m}"?
If so why self was the problem?
Another question: the code also tried to match "m" now it doesn't. Why?

os97673:
The changes looks clean now but I have a question (wrote it in the code)

lib/minitest/minitest_reporter_plugin.rb
adaedra: In the original code (`Runnable.run`), the method was called on the `Runnable` child class, so `self` was the name of the derived class and yields the desired result when assembling the test name.

Here, the `total_count` method is called on the Reporter, so `self` is the name of the reporter. Assembling the complete name uses wrongly the reporter name so matching on the test's parent class did not work.

It would also mean that if ran with a filter like `/Reporter/`, the progress bar would show more tests than actually run, because all tests would match.

The fix is to map the methods to use each runnable name when joining the test name so the filter works on the parent class name as intended.

If I add a `p` to show the assembled method name, on the `spec_test.rb` file:

Before: `"#<Minitest::Reporters::DelegateReporter:0x007f863e392058>#test_0001_works"`
After: `"String::#length#test_0001_works"`

For the second question, I considered that if `filter` matches `m`, it will also match `#{self}##{m}`, so both checks seemed redundant. Furthermore, it would require to return both test and full name from the `map` loops and the extra and redundant data. Here, we're collecting the minimal amount of data without changing the behavior.

os97673: Thank you for the explanation.  The change is more clear for me now.


> For the second question, I considered that if filter matches m, it will also match #{self}##{m}, so both checks seemed redundant. Furthermore, it would require to return both test and full name from the map loops and the extra and redundant data. Here, we're collecting the minimal amount of data without changing the behavior.

I think this will not be true if the patter try to match exact name (e.g. /^Pattern$/)

adaedra: That's correct, I forgot about this kind of filtering. Also, I was double wrong, as I don't need the bubble up the data so it doesn't add as much complexity as I initially thought.

I pushed a revision with the original filtering behavior. Let me know if you'd prefer have it squashed.

os97673:
Could you please add test for restrictive pattern we discussed (such as /^pattern$/) to make sure we will not break this functionality again.

os97673:
Thank you for the contribution, patience and the tests :)

PR #237, opened 2017-09-11
Only use DelegateReporter if it's been explicitly asked for.

pnomolos:
This will mean that the reporters are only overridden if `Minitest::Reporters.use!` has been called.

The `respond_to?` check is for cases where the plugin is on your GEM_PATH but you haven't actually required the gem.

### DIFF
--- a/lib/minitest/minitest_reporter_plugin.rb
+++ b/lib/minitest/minitest_reporter_plugin.rb
@@ -69,6 +69,7 @@ module Minitest
 
   class << self
     def plugin_minitest_reporter_init(options)
+      return unless Minitest::Reporters.respond_to?(:reporters) && Minitest::Reporters.reporters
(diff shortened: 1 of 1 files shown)


os97673:
interesting idea, but, perhaps, we should have empty `plugin_minitest_reporter_init` and rely on `use!`.  What do you think?

zenspider:
Either suggestion will address my concerns documented in #247. This is really needed as minitest-reporters breaks other plugins.

os97673:
I personally believe that there is no good way to avoid the breakage and provide a convenient behavior for users.  All this plugin idea doesn't work because some reporters naturally fight for the same medium to show the report (e.g. minitest itself enable a report which reports to a console and it means no other reporters can report to it :( )

So, I'm closing the PR and will think what is the best way to avoid such problem.

PR #238, opened 2017-09-11
Override Minitest.parallel_fork_stat_reporter if it's available

pnomolos:
- Provides integration with minitest-parallel_fork > 1.0.2

/cc @jeremyevans

### DIFF
--- a/lib/minitest/minitest_reporter_plugin.rb
+++ b/lib/minitest/minitest_reporter_plugin.rb
@@ -70,6 +70,12 @@ module Minitest
   class << self
     def plugin_minitest_reporter_init(options)
       reporter.reporters = [Minitest::Reporters::DelegateReporter.new(reporter.reporters, options)]
+
+      return unless respond_to?(:parallel_fork_stat_reporter)
+      define_singleton_method :parallel_fork_stat_reporter do |reporter|
+        reporter.reporters.detect { |rep| rep.is_a?(Minitest::Reporters::DelegateReporter) }
+                .send(:all_reporters).detect { |rep| rep.is_a?(StatisticsReporter) }
+      end
     end
   end
 end


os97673:
Sorry for late response: could you please clarify what is the purpose of the method you want to add?

pnomolos:
@os97673 See https://github.com/jeremyevans/minitest-parallel_fork/commit/a3c55aa3292c4357f8e1d0390c1a8f6613467236#diff-b9aaa6d358fee521899490d21e5475dbR43 - this will allow minitest-reporters to play nicely with minitest-parallel_fork :)

os97673:
Hmm, so you need some StatisticsReporter.  Am I right? If we have more than one such reporters which one we should return from the method.  More information on what you do expect would be helpful.

pnomolos:
`StatisticsReporter` is from minitest core - the default reporter.  Because minitest-parallel_fork checks for an `is_a?` as opposed to checking on method names (or similar) it currently doesn't play nicely with minitest-reporters (fails badly).  All I'm wanting to do here is to have minitest-parallel_fork properly pick up the minitest-reporters `DelegateReporter`

os97673:
I am not sure I understand what you actually want to achieve :(
Could you please describe what problem do you have, how we fail and how the
change will help to fix the problem you have.

The problem on my side that I only see the solution you implemented but do
not understand what it solves and why it solution and the code itself is
meaningless from minitest-reporters point of view :(

On Oct 2, 2017 4:59 PM, "Phil Schalm" <notifications@github.com> wrote:

StatisticsReporter is from minitest core - the default reporter. Because
minitest-parallel_fork checks for an is_a? as opposed to checking on method
names (or similar) it currently doesn't play nicely with minitest-reporters
(fails badly). All I'm wanting to do here is to have minitest-parallel_fork
properly pick up the minitest-reporters DelegateReporter

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/kern/minitest-reporters/pull/238#issuecomment-333699073>,
or mute the thread

os97673:
closing due to lack of response :(

PR #243, opened 2017-11-03
Use block syntax to close files after writing

rodreegez:
I added minitest-reporters to two large Rails apps in order to use
the MeanTimeReporter to identify the slowest running tests.

I started seeing `Errno::EMFILE: Too many open files` errors. The errors
occur intermittently locally, but consistently on CI. I don't really
think that it's minitest-reporters "fault", however when reviewing the
code I noticed that there were a couple places in the reporter that file
handles could be being left open.

This PR switches to using the block syntax to open and then write to a
file (as is done elsewhere in the reporter), which implicitly closes the
file when exiting the block.

### DIFF
--- a/lib/minitest/reporters/mean_time_reporter.rb
+++ b/lib/minitest/reporters/mean_time_reporter.rb
@@ -260,11 +260,15 @@ module Minitest
             previous_run.store("#{description}", new_times)
           end
 
-          File.write(previous_runs_filename, previous_run.to_yaml)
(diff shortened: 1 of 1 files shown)


os97673:
IO.write() doc (http://ruby-doc.org/core-2.4.2/IO.html#method-c-write)  says
"write ensures the file is closed before returning."
Why do you think we need to replace it with blocks?

PR #245, opened 2017-12-05
Make output in `single_file` mode use the proper suite, add `filepath` and `lineno` attribute

jules2689:
The problem
---
- single file mode prints out many suites in a single suite, but labels them all as "minitest"
- filepath/lineno, despite not being in spec, can be useful for ruby programs (and are standard in minitest/rails, for example) to find a test file to use in tests
- same with line number for a test

The changes
---
- single_file mode now prints a single file with a nested `testsuites` option (as per spec). The suite entries now all use the proper `suite` (see last point for example)
- filepath is calculated from the current directory (or the directory that you define with `:base_path`) and the filepath is added to the `testsuite` entry
- adds `lineno` to the `test` entry
- refactors slightly to adhere to [junit spec format](https://www.ibm.com/support/knowledgecenter/en/SSQ2R2_14.0.0/com.ibm.rsar.analysis.codereview.cobol.doc/topics/cac_useresults_junit.html):

```xml
<testsuites>
  <testsuite suite="A"></testsuite>
  <testsuite suite="B"></testsuite>
</testsuites>
```

### DIFF
--- a/lib/minitest/reporters/junit_reporter.rb
+++ b/lib/minitest/reporters/junit_reporter.rb
@@ -62,7 +62,8 @@ module Minitest
                       :errors => suite_result[:error_count], :tests => suite_result[:test_count],
                       :assertions => suite_result[:assertion_count], :time => suite_result[:time]) do
           tests.each do |test|
-            xml.testcase(:name => test.name, :classname => suite, :assertions => test.assertions,
+            lineno = test.method(test.name).source_location.last
+            xml.testcase(:name => test.name, :lineno => lineno, :classname => suite, :assertions => test.assertions,
                          :time => test.time) do
               xml << xml_message_for(test) unless test.passed?
             end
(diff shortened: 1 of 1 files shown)


jules2689:
@os97673 does this make sense to you? Should I ask for review from anyone else?

os97673:
> @os97673 does this make sense to you? Should I ask for review from anyone else?

Hi, sorry for late response (a little bit busy at work :(
Here are my thoughts:
- adding filepath and lineno is safe and may be useful for some people
- fixing suite name looks reasonable but some people may have issues with this (hope they will not)
- changing structure is also reasonable but most likely will affect some people 
Thus it would be great to have these changes separated so we could revert the structure change w/o reverting filepath and lineno if too many people will be affected by this.  
In any case I think the structure should be changed but if too many people affected we will need to add a config key to enable/disable it (but I do not want to add it too early to avoid redundant config keys)
Does this make sense for you?

os97673:
published as v1.2.0.beta2

jules2689:
Apologies @os97673 I have been out of office for a bit. I can still split this up for a full release if we want to revert?

os97673:
no problem, I've pushed it as is and published with assumption that you will help to fix problems later ;)

Issue #248, opened 2018-01-05
Minitest 5.11.1 regression for SpecReporter.


skunkworker:
On Minitest `5.10.3` spec reporter grouped the specs under the test file name,
like
``` 
FooTest
   test_bar       PASS
   test_ack       PASS
```
but after updating Minitest to `5.11.1`, they are now not grouped under the test file name, like
``` 
FooTest
   test_bar       PASS

FooTest
   test_ack       PASS
```

alexhayes:
I can confirm a similar issue for the RubyMine reporter and rolling back to minitest 5.10.3 fixes the 
issue.

![image](https://user-images.githubusercontent.com/142916/34845481-145037c4-f769-11e7-9a0d-18f4ee515aa0.png)

donnfelker:
I can also confirm. I'm having the same problem as @alexhayes in RubyMine. This question is also showing up on Stackoverflow now too. https://github.com/kern/minitest-reporters/issues/248

mbround18:
I can confirm this as well, I believe the issue stems from how [minitest](https://github.com/seattlerb/minitest) changed the contents of what gets passed into the `record` method.

sandstrom:
I think [this is the line](https://github.com/seattlerb/minitest/commit/00433fc0a4fdd0e6b302aace633384ba13122376#diff-efcebb0b0e7d5c021792f400fd891d9fR581) that prints it. But it seems like it has been there all along, so I can't really figure out why the issue occurred now.

os97673:
should be fixed by 53ef519fcad7a61c5bb4340be92b4697dca34739 and released as 1.2.0.beta1

mattbrictson:
Will 1.2.0 be released soon? 🙏

os97673:
published

tleish:
@os97673 - The problem might not be minitest-reporters, it looks like something changed with minitest, which is affecting the reporter and the way in which Rubymine uses it.  To test this, you can use minitest-reporters project. 

Running the minitest-reporters project inside Rubymine using _minitest 5.10.3_, the report outputs as expected.

```diff
- s.add_dependency 'minitest', '>= 5.0'
+ s.add_dependency 'minitest', '5.10.3'
```
minitest-reporters (1.3.0.beta2)

![image](https://user-images.githubusercontent.com/1099993/39540594-1ef9f3fa-4e11-11e8-91a6-f9ed670dece1.png)

[...]

PR #249, opened 2018-01-15
Check for nil value in before_test and modified check for calling after_suite in before_test

mbround18:
In response to #248 in the file `lib/minitest/reporters/base_reporter.rb` after the check for the nil the code was in place `after_suite` was still repetitively called and by adding the text check for `'Minitest::Result'` this will now call `after_suite` at the appropriate time. I tested on Minitest 5.11.1 using the SpecReporter, RubyMineReporter, DefaultReporter, and a custom reporter I am using. I followed that up by testing back on 5.10.3 to verify that it works on the older version as well.

### DIFF
--- a/lib/minitest/reporters/base_reporter.rb
+++ b/lib/minitest/reporters/base_reporter.rb
@@ -1,4 +1,3 @@
-require 'pp'
 module Minitest
   module Reporters
     class BaseReporter < Minitest::StatisticsReporter
@@ -15,7 +14,7 @@ module Minitest
 
       # called by our own before hooks
       def before_test(test)
-      	last_test = tests.last
+        last_test = tests.last
         before_suite(test.class) if last_test.nil?
         after_suite(last_test.class) if last_test == 'Minitest::Result'
       end
--- a/lib/minitest/reporters/base_reporter.rb
+++ b/lib/minitest/reporters/base_reporter.rb
@@ -1,3 +1,4 @@
+require 'pp'
 module Minitest
   module Reporters
     class BaseReporter < Minitest::StatisticsReporter
@@ -14,11 +15,9 @@ module Minitest
 
       # called by our own before hooks
       def before_test(test)
-        last_test = tests.last
-        if last_test.class != test.class
-          after_suite(last_test.class) if last_test
-          before_suite(test.class)
-        end
+      	last_test = tests.last
+        before_suite(test.class) if last_test.nil?
+        after_suite(last_test.class) if last_test == 'Minitest::Result'
       end
 
       def record(test)
--- a/lib/minitest/reporters/base_reporter.rb
+++ b/lib/minitest/reporters/base_reporter.rb
@@ -15,8 +15,11 @@ module Minitest
       # called by our own before hooks
       def before_test(test)
         last_test = tests.last
-        before_suite(test.class) if last_test.nil?
-        after_suite(last_test.class) if last_test == 'Minitest::Result'
+        if last_test.nil?
+          before_suite(test.class)
+        elsif is_a?(last_test.class) == String
+          after_suite(last_test.class)
+        end
       end
 
       def record(test)


lib/minitest/reporters/base_reporter.rb
```diff
@@ -1,3 +1,4 @@
+require 'pp'
```
os97673: redundant require?

lib/minitest/reporters/base_reporter.rb
```diff
@@ -14,11 +15,9 @@ def add_defaults(defaults)
 
       # called by our own before hooks
       def before_test(test)
-        last_test = tests.last
-        if last_test.class != test.class
-          after_suite(last_test.class) if last_test
-          before_suite(test.class)
-        end
+      	last_test = tests.last
```
os97673: it looks like indentation is broken

os97673:
to be able to review the changes it would be helpful if you'd describe the cause of the problem w/o it the diff looks like a collection of some random changes which happen to produce desired outcome :(

sandstrom:
@os97673 I think what's happened here is that in [`00433fc0`](https://github.com/seattlerb/minitest/commit/00433fc0a4fdd0e6b302aace633384ba13122376#diff-efcebb0b0e7d5c021792f400fd891d9fR930) a change was introduced where `run` would return an instance of `Result`. 

Because of this change the conditional `if last_test.class != test.class` (in this diff/PR) will never be true, so the `before_suite` callback is called before every test.

Does that make sense?

*(I agree with you that `pp` shouldn't be there!)*

lib/minitest/reporters/base_reporter.rb
mbround18: Fixed indentation

mbround18:
I added a commit which removes the redundant require. (I was using that for debugging and didn't mean to commit it) I also fixed the indentation as per comment. The reason for the change is the logic behind the initial code calls `before_suite` repetitively so in other reporters that do something like puts or streams the test class name to IO causes the before to be called every time. Next I made a nil check because when I was running the code locally tests.last does not initially exist because it is an empty array to begin with. Following that I noticed that `after_suite` was still being called after every test; So with a little debugging using `pp` I looked into what was happening and on the last test in the suite the object for `Minitest::Result` had not yet become an invocation of the actual class which is why I added the simple string check on the final test so it will than call `after_suite`.

os97673:
Thank you for the explanation.  Need to think about it.  Still have some questions: e.g. why we call before_suite() and after_suite() in different calls? or why this code works for older versions of minitest.

mbround18:
@os97673 I looked into this a bit and when you are on minitest version `>= 5.10.3` and modify master in the `BaseReporter` to include the following code and run a test suite with a couple test in it (in the return I added comments note that the comments wouldnt be there if you ran it yourself :P ):
```ruby
      # called by our own before hooks
      def before_test(test)
        last_test = tests.last
        pp last_test.class
        pp test.class
        if last_test.class != test.class
          after_suite(last_test.class) if last_test
          before_suite(test.class)
        end
      end
```
It will return with:
```ruby
NilClass # => run on TestSuite#test_001
TestSuite  # => run on TestSuite#test_001
TestSuite  # => run on TestSuite#test_002
TestSuite  # => run on TestSuite#test_002
TestSuite  # => run on TestSuite#test_003
TestSuite  # => run on TestSuite#test_003
TestSuite  # => run on TestSuite#test_004
TestSuite  # => run on TestSuite#test_004
```

However when you update to minitest `<= 5.11.0` the return will be:
```ruby 
NilClass # => run on TestSuite#test_001
TestSuite # => run on TestSuite#test_001
Minitest::Result # => run on TestSuite#test_002
TestSuite # => run on TestSuite#test_002
Minitest::Result # => run on TestSuite#test_003
TestSuite # => run on TestSuite#test_003
Minitest::Result # => run on TestSuite#test_004
TestSuite # => run on TestSuite#test_004
```


The culprit for the the issue lies within the `MiniTest::Test`  in the commit https://github.com/seattlerb/minitest/commit/00433fc0a4fdd0e6b302aace633384ba13122376#diff-444fa8babd8886c012733b064d709db5 when they changed how they return in the method `run` from being an instance of itself (itself being the test class as seen in the first result) to the return being an instance of the new class called `Result` see below, the change occured on line 104 of the `test.rb` file in the minitest code repository: 
```ruby
-        self # per contract
+       Result.from self # per contract
```

Side note I have a bit of a code change which leans up my code a bit so the code base wouldnt be dependent on minitest maintaining the name `MiniTest::Result` see my following commit.

lib/minitest/reporters/base_reporter.rb
```diff
@@ -15,9 +15,10 @@ def add_defaults(defaults)
       # called by our own before hooks
       def before_test(test)
         last_test = tests.last
-        if last_test.class != test.class
-          after_suite(last_test.class) if last_test
+        if last_test.nil?
           before_suite(test.class)
+        elsif is_a?(last_test.class) == String
```
sandstrom: Minor comment, but how about `elsif last_test.class.is_a?(String)` here?

sandstrom:
@os97673 I think @mbround18 explained it well, but otherwise let me know and I'll try to explain to the best of my ability.

mbround18:
I worked with @aarontc on looking to the issue and him and I worked out a better solution to this issue, I am closing this pull request and am going to open a new one with this one as a reference.

PR #250, opened 2018-01-23
Adding test for response to klass on MiniTest::Result object

mbround18:
This is in reference to https://github.com/kern/minitest-reporters/pull/249 the people in the https://github.com/seattlerb/minitest project made an API change with version 5.11.3+ on the commit https://github.com/seattlerb/minitest/commit/00433fc0a4fdd0e6b302aace633384ba13122376#diff-444fa8babd8886c012733b064d709db5 where they added the object Result. The solution I am proposing with collaboration from @aarontc is that a check performed on `last_test` to respond to the variable `klass` of which is where minitest now stores the class name of the test class. After this check is performed a not equal is checked in both places of the if else statement to validate the start of a new suite. I intend to log an issue with the people at minitest around the API level change for a minor version.

### DIFF
--- a/lib/minitest/reporters/base_reporter.rb
+++ b/lib/minitest/reporters/base_reporter.rb
@@ -1,4 +1,3 @@
-require 'pp'
 module Minitest
   module Reporters
     class BaseReporter < Minitest::StatisticsReporter
--- a/lib/minitest/reporters/base_reporter.rb
+++ b/lib/minitest/reporters/base_reporter.rb
@@ -1,3 +1,4 @@
+require 'pp'
 module Minitest
   module Reporters
     class BaseReporter < Minitest::StatisticsReporter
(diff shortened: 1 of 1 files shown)


mbround18:
I added an issue surrounding this over on the minitest repository https://github.com/seattlerb/minitest/issues/738

lib/minitest/reporters/base_reporter.rb
```diff
@@ -1,3 +1,4 @@
+require 'pp'
```
mbround18: I removed this in the second commit.

sandstrom:
@mbround18 Awesome fix!

The indentation look a bit weird on my display (photo below). Could that be a tabs vs. spaces issue? Perhaps you can try converting them to one or the other and push the changes here (to see if it goes away).

image

sandstrom:
@os97673 do you have any feedback on this PR?

os97673:
I've cleaned the change (it looks like it had some tabs), merged it as 53ef519fcad7a61c5bb4340be92b4697dca34739, and released as 1.2.0.beta1

PR #252, opened 2018-02-03
Adapt JUnitReporter for minitest >= 5.11

Kevinrob:
Minitest 5.11.0 change the format of `test` passed to `record`.
`test.class` return `Minitest::Result`, we have to call `test.klass`.

The issue is that JUnit parser will think that it's the same test if the `name` and `classname` is the same.


Kevinrob:
**Correct output:**  
correct
  
**Not correct output:**  
issue

PR #255, opened 2018-02-28
Allow force of specified reporter when calling use!

brettwgreen:
- Had some trouble getting TeamCity to run JUnit style tests. Took some digging to discover this TeamCity detection that forces it to use RubyMineReporter regardless of what I pass to `use!`
- I'd like to use the JUnit style tests, while retaining the flexibility of using other reporters locally. I am runnning command line based unit tests in TeamCity to avoid the limitations of the Rake Runner build step, which does not play well with multiple ruby versions across many projects. Using JUnitReporter in that context allows me to collect the XML reports in TeamCIty and use the natural goodness of rvm and bundler.
- This change allows users to pass a force flag and ignore environment variables and just use the runner that was asked for while retaining backwards compatibility.

### DIFF
--- a/lib/minitest/reporters.rb
+++ b/lib/minitest/reporters.rb
@@ -36,7 +36,7 @@ module Minitest
       end
     end
 
-    def self.use_runner!(console_reporters, env, force = false)
+    def self.use_runner!(console_reporters, env, force)
       self.reporters = choose_reporters(console_reporters, env, force)
     end
 
@@ -58,7 +58,7 @@ module Minitest
       end
     end
 
-    def self.choose_reporters(console_reporters, env, force = false)
+    def self.choose_reporters(console_reporters, env, force)
       if env["TM_PID"] && !force
         [RubyMateReporter.new]
       elsif (env["RM_INFO"] || env["TEAMCITY_VERSION"]) && !force
--- a/lib/minitest/reporters.rb
+++ b/lib/minitest/reporters.rb
@@ -22,7 +22,7 @@ module Minitest
       attr_accessor :reporters
     end
 
-    def self.use!(console_reporters = ProgressReporter.new, env = ENV, backtrace_filter = nil)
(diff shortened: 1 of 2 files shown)


brettwgreen:
Don't know why that test would fail given my change. It works locally. I suspect the test is flaky somehow, or some kind of environmental difference?

brettwgreen:
Ahh... I see. It only failed on jruby-9.1.9.0. Still don't think my change would have introduced a problem related to sorting test output.

os97673:
The original idea was to always use RubyMineReporter for RubyMine and TeamCity even if user specified this otherwise (in the code), so RM and TeamCity could format reports.
The suggested change allows to write code which will break the original idea.  I'd prefer to have special environment variable which would allow to achieve your goal (e.g. we could ignore all env. vars in `choose_reporter`) this way we will preserve the original intent and provide the required functionality. WDYT? 
I'm closing the PR as a sign that I will not accept it in the current form (unless you convince me ;)  But I'm open for discussions.

brettwgreen:
That's fine... In the end I ended up doing something similar to what you suggest... i.e. using REPORTER env variable and using that as reporter when present along with passing an empty hash for env which circumvents the teamcity detection.

os97673:
Then I'd suggest to change choose_reporter to check this env var first
(perhaps MINITEST_REPORTER would be better name). Are you willing/have time
to do such PR?

Thanks in advance,  Oleg.

On Mar 14, 2018 2:12 PM, "Brett W Green" <notifications@github.com> wrote:

That's fine... In the end I ended up doing something similar to what you
suggest... i.e. using REPORTER env variable and using that as reporter when
present along with passing an empty hash for env which circumvents the
teamcity detection.

—
You are receiving this because you modified the open/close state.

Reply to this email directly, view it on GitHub
<https://github.com/kern/minitest-reporters/pull/255#issuecomment-373176301>,
or mute the thread

brettwgreen:
Sure... I will take a look sometime this week.

Issue #257, opened 2018-03-23
Latest JUnitReporter seems to break JUnit spec


brettwgreen:
Root element of JUnitReporter's XML files seems to have changed from `<testsuites>` to `<test_suites>` which appears to be in violation of the JUnit/Ant spec

Some looking around seems to confirm that `<testsuites>` is the root element spec, and it certainly is rejected now by TeamCity's XML Report Processer -> Ant JUnit processor. It now gives this error:

> File test/reports/TEST-Foo.xml doesn't match the expected format: "testsuites" or "testsuite" root element expected

So the PR that was merged (https://github.com/kern/minitest-reporters/pull/256) gave me integrated JUnit reporter override capability, but broke my ability to use it effectively after upgrade broke the XML format.

If you agree that the spec has been broken and should be fixed, happy to put a PR in to try and fix/test it against the spec.

Seems to have been modified with this commit:
https://github.com/kern/minitest-reporters/commit/e120409fb67363b98047997e590da43d1946f0fb

Some specs here:
http://junitpdfreport.sourceforge.net/managedcontent/PdfTranslation
http://help.catchsoftware.com/display/ET/JUnit+Format

os97673:
yes, I agree that it should be `<testsuites>`  PR is welcome.

os97673:
fixed by #260 260

brettwgreen:
This is functioning as desired on my TeamCity server now when enabling the JunitReporter with the new ENV variable. and using XML report collection with now proper JUnit spec'd XML. Just thought I'd leave the feedback.

os97673:
@brettwgreen let me know if you find any problems with the latest version.

PR #260, opened 2018-04-14
Fix Junit Spec

brettwgreen:
As referenced by: https://github.com/kern/minitest-reporters/issues/257


os97673:

PR #265, opened 2018-06-13
Fix/time reporters

brendandeere:
### Problem
The time based reporters stopped reporting correctly when `Minitest::Result` was introduced.

### Fix
Parse the Test and MiniTest::Result test class using the #test_class helper method.

### Tested On
Ruby 1.9.3
Ruby 2.3.1 
Ruby 2.5.1

### Steps to reproduce
https://gist.github.com/brendandeere/f8b9788f9d273298adbf64ddf79d758b

### Before:

screen shot 2018-06-12 at 7 50 51 pm

### After
screen shot 2018-06-12 at 7 51 24 pm

PR #266, opened 2018-06-19
Invoke  "after_suite()" on a 'klass' instead of 'class' for new 5.11 API

kradima:
Since 5.11 'class' gets 'Minitest::Result' object instead of actual suite name. Let's use 'klass' for `after_suite()` if possible.

### DIFF
--- a/lib/minitest/reporters/base_reporter.rb
+++ b/lib/minitest/reporters/base_reporter.rb
@@ -17,14 +17,15 @@ module Minitest
         last_test = tests.last
 
         # Minitest broke API between 5.10 and 5.11 this gets around Result object
-        if last_test.respond_to? :klass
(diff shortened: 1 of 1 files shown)


kradima:
Since 5.11 RubyMine has stopped to show tests as tree. The reason is that `last_test.class` hasn't got actual test class name but `klass` should be used instead.

@os97673 Could you please take a look around?

os97673:
sorry for late response. it looks like your change is trying to address the same problem as #265 and the solution suggested in #265 looks better (but doesn't work for 1.9.3)  So, it would be great if you would help to fix it for 1.9.3 (or explain why your solution is better ;)

donv:
1.3.1.beta1 seems to solve the problem in RubyMine for me.  Thanks!

PR #269, opened 2018-08-31
Fix mean time reporter, broken since 1.3.2

duonoid:
I believe a small bug was introduced in 271141c2aaf26ed60db42e9daf57a88dcf35ac66, which seems to be fixed with this patch.

I did not run the test on Windows.


os97673:
thank you for fixing the problem

Issue #271, opened 2018-10-08
After running `rake reset_statistics`, Minitest::Reporters::MeanTimeReporter breaks Minitest


jdickey:
The [`:reset_statistics` Rake task](https://github.com/kern/minitest-reporters/blob/master/Rakefile#L56-L61) from the project Rakefile is mentioned numerous times in the source for [`Minitest::Reporters::MeanTimeReporter`](https://github.com/kern/minitest-reporters/blob/master/lib/minitest/reporters/mean_time_reporter.rb) as doing what it says, and the Rake task itself looks too simple to fail.

However, after running the Rake task and then running `rake`, I see the following error output:

```
gemset/gems/minitest-reporters-1.3.5/lib/minitest/reporters/mean_time_reporter.rb:251:in `block in create_or_update_previous_runs!': undefined method `[]' for false:FalseClass (NoMethodError)
	from .../minitest-reporters-1.3.5/lib/minitest/reporters/mean_time_reporter.rb:250:in `each'
	from .../minitest-reporters-1.3.5/lib/minitest/reporters/mean_time_reporter.rb:250:in `create_or_update_previous_runs!'
	from .../minitest-reporters-1.3.5/lib/minitest/reporters/mean_time_reporter.rb:79:in `report'
	from .../minitest-reporters-1.3.5/lib/minitest/minitest_reporter_plugin.rb:26:in `each'
	from .../minitest-reporters-1.3.5/lib/minitest/minitest_reporter_plugin.rb:26:in `report'
	from .../minitest-5.11.3/lib/minitest.rb:808:in `each'
	from .../minitest-5.11.3/lib/minitest.rb:808:in `report'
	from .../minitest-5.11.3/lib/minitest.rb:141:in `run'
	from .../minitest-tagz-1.6.0/lib/minitest/tagz.rb:56:in `run'
	from .../minitest-5.11.3/lib/minitest.rb:63:in `block in autorun'

# Running tests with run options --seed 62125:

.........

Finished tests in 0.002525s, 3564.3564 tests/s, 3564.3564 assertions/s.


9 tests, 9 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for Unit Tests to .../tmp/coverage. 26 / 26 LOC (100.0%) covered.
rake aborted!
Command failed with status (1)

Tasks: TOP => default => test
(See full trace by running task with --trace)
```

If I run the specs using *any other reporter*, there's no problem.

[...]

PR #273, opened 2018-11-04
Fix MeanTimeReporter#reset_statistics!, add specs and cleanup

jdickey:
As noted in #271, [`Minitest::Reporters::MeanTimeReporter#reset_statistics!`](https://github.com/kern/minitest-reporters/blob/master/lib/minitest/reporters/mean_time_reporter.rb#L103) is inconsistent with the test in [`Minitest::Reporters::MeanTimeReporter#previously_ran?`](https://github.com/kern/minitest-reporters/blob/master/lib/minitest/reporters/mean_time_reporter.rb#L216), and that crashes each future Minitest run until the previous-runs file is manually deleted. This changes the `#reset_statistics` method to delete the file instead of overwriting it with an empty file, as is presently done.

Before that commit (#edb7e2d), we replace [the empty `MeanTimeReporterTest` class](https://github.com/kern/minitest-reporters/blob/d28c3ad9630234a306359534191ad4c8ba35bbfc/test/integration/reporters/mean_time_reporter_test.rb) with a class containing tests proving that all constructor options to `MeanTimeReporter` work as expected.

Finally, we make first `MeanTimeReporter`, then all other source files in `lib` RuboCop-clean, either by repairing the offending lines of code or by wrapping them in comments which disable the relevent "cops" over the offending lines. We also modify the `Rakefile` to run the `rubocop` task after successful completion of the `test` task by default, so that later changes to *files in the `lib` directory only* will have those changes inspected by RuboCop. The `Rakefile` *does not* presently run RuboCop against test files; we **recommend** that files matching `test/**/*_test.rb` be made RuboCop-clean in a future PR, with the `Rakefile` properly adjusted at that time.

### DIFF
--- a/lib/minitest/reporters/html_reporter.rb
+++ b/lib/minitest/reporters/html_reporter.rb
@@ -45,6 +45,7 @@ module Minitest
       def friendly_name(test)
         groups = test.name.scan(/(test_\d+_)(.*)/i)
         return test.name if groups.empty?
+
         "it #{groups[0][1]}"
       end
 
@@ -57,6 +58,7 @@ module Minitest
       def initialize(args = {})
         super({})
 
+        # rubocop:disable Layout/AlignHash
         defaults = {
           :title           => 'Test Results',
           :erb_template    => "#{File.dirname(__FILE__)}/../templates/index.html.erb",
@@ -64,6 +66,7 @@ module Minitest
           :mode            => :safe,
           :output_filename => 'index.html',
         }
+        # rubocop:enable Layout/AlignHash
 
         settings = defaults.merge(args)
 
@@ -205,6 +208,7 @@ module Minitest
         last_before_assertion = ''
         exception.backtrace.reverse_each do |s|
           break if s =~ /in .(assert|refute|flunk|pass|fail|raise|must|wont)/
+
           last_before_assertion = s
         end
         last_before_assertion.sub(/:in .*$/, '')
--- a/lib/minitest/reporters/html_reporter.rb
+++ b/lib/minitest/reporters/html_reporter.rb
@@ -45,7 +45,6 @@ module Minitest
       def friendly_name(test)
         groups = test.name.scan(/(test_\d+_)(.*)/i)
         return test.name if groups.empty?
-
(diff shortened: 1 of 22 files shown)


jdickey:
Now works with all 2.x Rubies. 1.9.3 is going to take a bit more work since many of the Gems (e.g., `rake`) have current versions that require Ruby 2+; this becomes more true the farther away we get from the 1.9/2.0 chasm.

lib/minitest/reporters/html_reporter.rb
```diff
@@ -57,13 +58,15 @@ def friendly_name(test)
       def initialize(args = {})
         super({})
 
+        # rubocop:disable Layout/AlignHash
```
os97673: Why do we need this?

lib/minitest/reporters/junit_reporter.rb
```diff
@@ -155,7 +156,9 @@ def filename_for(suite)
         while File.exist?(File.join(@reports_path, filename)) # restrict number of tries, to avoid infinite loops
           file_counter += 1
           filename = "TEST-#{suite_name}-#{file_counter}.xml"
+          # rubocop:disable Style/AndOr
```
os97673: Why do we need this?

lib/minitest/reporters/progress_reporter.rb
```diff
@@ -19,6 +19,7 @@ def initialize(options = {})
         super
         @detailed_skip = options.fetch(:detailed_skip, true)
 
+        # rubocop:disable Layout/AlignHash
```
os97673: same here.  I'd prefer to avoid such directives unless we really have to.

lib/minitest/reporters/progress_reporter.rb
```diff
@@ -75,7 +78,7 @@ def report
       private
 
       def show
-        @progress.increment unless count == 0
+        @progress.increment unless count.zero?
```
os97673: Could you please point me to the explanation why it is better.

os97673:
no response, so closing the PR for now.  Feel free to reopen with answers/fixes

PR #277, opened 2019-01-21
Add ability to set reports directory via environment variable

KevinSjoberg:
This pull request add the ability to specify the reports directory via an environment variable. This is especially useful when setting the given reporter via the environment variable `MINITEST_REPORTER`.

Further, we're currently using Bitbucket Pipeline for CI purposes. They enforce certain expectations on where the Xunit-compatible output is located. Currently, we check for the `CI` environment variable in our `test_helper.rb` and change the reporter and the expected reports directory as needed. With this change, we could just set `MINITEST_REPORTER` and `MINITEST_REPORTERS_REPORTS_DIR` in our Bitbucket environment and remove any custom changes made explicitly for Bitbucket from our `test_helper.rb`.


os97673:
Thank you for the contribution.  Sorry for late response.  Hope to publish a new version in a week.

PR #280, opened 2019-06-28
Normalise friendly name usage

rurounijones:
Normalize the use of `friendly_name` between the HTML and JUnit
reporters when reporting results for `Minitest::Spec`

#### Note
HTML and JUnit are not identical because I kept the HTML "It" prefix, none of the JUnit parsers / reporter systems I have seen display test names in a way that would have benefited from the "it".

I only use the HTML and JUnit reporters so I am not sure if any of the others would benefit from being normalised as well.

### DIFF
--- a/lib/minitest/reporters/base_reporter.rb
+++ b/lib/minitest/reporters/base_reporter.rb
@@ -46,6 +46,13 @@ module Minitest
         before_suite(test_class(test))
       end
 
+      # Trims off the number prefix on test names when using Minitest Specs
+      def friendly_name(test)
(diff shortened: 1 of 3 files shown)


os97673:
Could you please add tests for this

rurounijones:
Could you expand on your expectations of tests here? I looked at the existing tests for those files before committing and there were none that did this kind of full input/output testing.

os97673:
I'd imagine something similar to test/unit/minitest/spec_reporter_test.rb (create a dummy test object, call report and make sure the right name has been reported)

rurounijones:
Went with a custom set of reporters, closing issue.

PR #286, opened 2019-08-19
Use exception class for error, failure and skipped type

### LINKED ISSUE #285: junit.xml uses the wrong value for the error type field
`Minitest::Reporters::JUnitReporter`'s output has the test name for the `skip`, `error` and `failure` elements' `type` attribute, which doesn't seem to make sense.  It is also redundant, since we already have the test name in the parent `testcase` element's `name` attribute.

[Apache Ant's JUnit XML schema description](https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd) describes the error's type as

> The type of error that occured. e.g., if a java execption is thrown the full class name of the exception.

so it should clearly be the error class.

The failure type is described as

> The type of the assert

which seems more vague, but still clear enough that it isn't referring to the test name.

The `skipped` element isn't even a part of their junit.xml schema.

dylanahsmith:
Fixes #285

## Problem

Using the test name for the error or failure type didn't make sense, wasn't aligned with the [Apache Ant's JUnit XML schema description](https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd) of the error type and wasn't useful by redundantly including the test name again in this field.

## Solution

Use the exception class name for the type of the skipped, failure and error elements.

### DIFF
--- a/lib/minitest/reporters/junit_reporter.rb
+++ b/lib/minitest/reporters/junit_reporter.rb
@@ -99,16 +99,16 @@ module Minitest
           txt.sub(/\n.*/m, '...')
         end
 
-        e = test.failure
+        failure = test.failure
 
         if test.skipped?
-          xml.skipped(:type => test.name)
+          xml.skipped(:type => failure.error.class.name)
         elsif test.error?
-          xml.error(:type => test.name, :message => xml.trunc!(e.message)) do
+          xml.error(:type => failure.error.class.name, :message => xml.trunc!(failure.message)) do
             xml.text!(message_for(test))
           end
-        elsif test.failure
-          xml.failure(:type => test.name, :message => xml.trunc!(e.message)) do
+        elsif failure
+          xml.failure(:type => failure.error.class.name, :message => xml.trunc!(failure.message)) do
             xml.text!(message_for(test))
           end
         end


lib/minitest/reporters/junit_reporter.rb
```diff
@@ -99,16 +99,16 @@ def xml.trunc!(txt)
           txt.sub(/\n.*/m, '...')
         end
 
-        e = test.failure
+        failure = test.failure
 
         if test.skipped?
-          xml.skipped(:type => test.name)
+          xml.skipped(:type => failure.error.class.name)
```
os97673: Could you please point in the spec where this is specified?

os97673:
The changes looks reasonable except the one for skipped: I've failed to find this node in the spec and name of the test looks more useful information there.

lib/minitest/reporters/junit_reporter.rb
dylanahsmith: The skipped element doesn't appear to be in the spec.

> name of the test looks more useful information there.

I don't see why it is useful, since it is in the test case name already.  It just seemed like it made more sense to keep it consistent with the error and failure elements.

PR #287, opened 2019-08-25
Added date and time to the HTMLReporter

cderche:
We produce reports that we need to show customers. Having the date and time automatically inserted in the report is a big win :)

### DIFF
--- a/lib/minitest/templates/index.html.erb
+++ b/lib/minitest/templates/index.html.erb
@@ -11,8 +11,8 @@
 <body>
 <div class="container">
     <div class="jumbotron">
+        <p class="text-muted">Generated on <%= DateTime.now.strftime("%b %d, %Y at %H:%M") %></p>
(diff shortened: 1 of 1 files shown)


lib/minitest/templates/index.html.erb
```diff
@@ -11,7 +11,9 @@
 <body>
 <div class="container">
     <div class="jumbotron">
-        <h1><%= title %></h1>
+        <h3><small>Generated on <%= DateTime.now.strftime("%b %d, %Y at %H:%M") %></small></h3>
```
os97673: Are you sure you don't need to specify time zone the reporter should use?  For generic change I'd expect it should work correctly if the reviewer and the reporter creator are in different time zones.

cderche: Good idea! Done!

Issue #290, opened 2019-09-23
BaseReporter fails to get name of the test case in `before_test` handler


avokin:
The issue caused by `base_reporter.rb`. It takes nil from `result.klass` even if name of test could be got from `result.class.name`

barrettkingram:
Closing this - see thread in #291

PR #291, opened 2019-09-23
Correctly retrieve name of the test case when case#klass is nil

avokin:
BaseReporter fails to get name of the test case in `before_test` handler #290

In ActionRecord TestCase attribute `klass` could return nil. That's why it necessary to try get name of the class from `class.name`

### DIFF
--- a/lib/minitest/reporters/base_reporter.rb
+++ b/lib/minitest/reporters/base_reporter.rb
@@ -82,12 +82,16 @@ module Minitest
       def test_class(result)
         # Minitest broke API between 5.10 and 5.11 this gets around Result object
         if result.nil?
-          nil
-        elsif result.respond_to? :klass
-          Suite.new(result.klass)
-        else
-          Suite.new(result.class.name)
+          return nil
         end
+        result_class = nil
+        if result.respond_to? :klass
+          result_class = result.klass
+        end
+        if result_class.nil?
+          result_class = result.class.name
+        end
+        Suite.new(result_class)
       end
 
       def print_colored_status(test)
--- a/test/unit/minitest/spec_reporter_test.rb
+++ b/test/unit/minitest/spec_reporter_test.rb
@@ -57,6 +57,14 @@ module MinitestReportersTest
 
       refute error_msg, error_msg
     end
+
+    def test_suite_class_retrieval
+      Object.const_set('MockingTest', Class.new { def klass() nil end})
+      @reporter.io = StringIO.new
+      @reporter.before_test(MockingTest.new)
+
+      assert_match /MockingTest/, @reporter.io.string
+    end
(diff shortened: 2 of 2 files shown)


avokin:
Issue in RubyMine tracker https://youtrack.jetbrains.com/issue/RUBY-24926

avokin:
My test doesn't work on Travis. I'm investigating it

avokin:
I have to remove integration test for RubyMineReporter, because it requires some files from RubyMine distributive. I've introduced a unit test instead.

os97673:
The code itself looks reasonable but have you figure out why klass returns nil?
Based on https://github.com/seattlerb/minitest/blob/master/lib/minitest.rb#L509 klass is the correct way to get name of the result.

avokin:
In Rails it's possible to create spec-style tests
`require 'test_helper'
require 'minitest/spec'

class MyEntityTest < ActiveSupport::TestCase
  describe 'Entity' do
    it 'should do something' do
    end
  end
end
`

This generates dynamic class for 'Entity', see Rails method `ActiveSupport::Testing::Declarative::extended`

Dynamically created class responds to :klass, but returns nil

os97673:
how does minitest works in this case?

os97673:
> how does minitest works in this case?

I'm asking this to figure out if this is a problem in minitest-reporters or in rails.  If minitest also expects klass to be the source of the test name then it is a problem in the rails' implementation

avokin:
minitest without minitest-reporters (for example command `rake tests`) works fine.

barrettkingram:
Is this still an issue? I'm unable to reproduce an example where `result.klass` is undefined with the spec-style tests.

Example:
```ruby
require "minitest/autorun"
require "minitest/reporters"
Minitest::Reporters.use!

class Foo
  def foo
    "foo"
  end
end

describe Foo do
  it "good test" do
    assert_equal "foo", Foo.new.foo
  end

  it "bad test" do
    assert_equal "bar", Foo.new.foo
  end
end
```
This produces output:
```
Started with run options --seed 48532

 FAIL Foo#test_0002_bad test (0.00s)
        Expected: "bar"
          Actual: "foo"
        simple_spec.rb:17:in `block (2 levels) in <main>'

  2/2: [==========================================================================] 100% Time: 00:00:00, Time: 00:00:00

Finished in 0.00112s
2 tests, 2 assertions, 1 failures, 0 errors, 0 skips
```
I also tried a similar example in RubyMine and wasn't able to reproduce the issue described in the RubyMine tracker: https://youtrack.jetbrains.com/issue/RUBY-24926

barrettkingram:
Closing for now but happy to reconsider if anyone can confirm this is still needed.

PR #301, opened 2020-04-16
Remove debug output from ProgressReporter

wvanbergen:
I noticed that the ProgressReporter was debugging details about failed tests twice: one time as  inspected array, then as proper string. Also, the class name and test name were reversed. I suspect the inspected array was a debug statement that accidentally was left in place.

I removed the debug print statement, and fixed the order of the test class and method name to be `Suite#test_method`.

### DIFF
--- a/lib/minitest/reporters/progress_reporter.rb
+++ b/lib/minitest/reporters/progress_reporter.rb
@@ -87,8 +87,7 @@ module Minitest
       end
 
       def print_test_with_time(test)
-        puts [test.name, test_class(test), total_time].inspect
-        print(" %s#%s (%.2fs)" % [test.name, test_class(test), total_time])
+        print(" %s#%s (%.2fs)" % [test_class(test), test.name, total_time])
       end
 
       def color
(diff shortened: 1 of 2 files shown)


wvanbergen:
This is what my output was looking like:

```
 FAIL["test_capture", #<Minitest::Reporters::Suite:0x00007ff1ffc9b578 @name="ShopifyBuild::Command2Test">, 0.062443000031635165]
 test_capture#ShopifyBuild::Command2Test (0.06s)
```

After this fix:

```
 FAIL ShopifyBuild::Command2Test#test_capture (0.05s)
```

itsliamegan:
I think this is good for a merge. After digging through the git-blame for the progress reporter, it seems like [this](https://github.com/kern/minitest-reporters/commit/4a15b7b88a3e921225a8e16dcd39684d6e2c6f09#diff-ddffb4844a251095ab26b2450f561773a5a1b563fb1b8f50dc00f01ced58eca1R92) was the commit that introduced it.

Seeing as it doesn't make sense to print the literal array, and it was introduced when doing some pretty major changes to the ProgressReporter, I'd have to agree that it was probably a debug statement accidentally left in that we can get rid of.

itsliamegan:
The CI failure also doesn't seem to have anything to do with this issue. My guess is that it was an intermittent failure, as CI sometimes does 🤷

os97673:
@itsliamegan hope you will update changelog.md for this too ;-)

PR #302, opened 2020-06-30
Add timestamps to changelog

sandstrom:
Easier to determine release cadence etc.

Not adding timestamps for all old releases, only for the past few ones. Should be enough.

Thanks for a good gem! 💎


os97673:

PR #310, opened 2021-03-08
fix a mispelling in README

deronnax:
I think the author meant "TextMate" and no "RubyMate", which I believe is a mixup between RubyMine and TextMate


itsliamegan:
Great find! Confusingly, it seems as if "RubyMate" is the term used throughout the codebase; since changing all references to "RubyMate" would require changing a class name (which would be a massive breaking change), I think this is sufficient to aid clarity in the README.

PR #311, opened 2021-04-04
Allow specifying the output folder and file name via env variables

estebanbouza:
Allow specifying the output folder and file name via env variables for the HTML reporter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ### [dev](https://github.com/kern/minitest-reporters/compare/v1.4.3...master)
 
+* Added support for environment variables to define the output location of HTML reports. [#311](https://github.com/minitest-reporters/minitest-reporters/pull/311) contributed by [estebanbouza](https://github.com/estebanbouza)
+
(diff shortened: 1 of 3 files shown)


itsliamegan:
Thanks for your contribution, @estebanbouza!

Just a couple notes:
1. Each section in the changelog indicates what changed between one version and another. That means that the "dev" section contains changes made since the last released version, and the numbered sections indicate what changed between each version. 
2. The way you've changed it, now it looks like the newest version is 1.4.4 and the changes from 1.4.2 to 1.4.3 are actually from 1.4.2 to 1.4.4, and 1.4.3 doesn't exist. It's a bit of a confusing file with all the markup, so I understand.

I don't think this is a major enough change to release a new version; with that in mind, here are the changes I'd like you to make:
1. Revert your changes to the version numbers in the changelog and `version.rb` file
2. Add a note under the "dev" section of the changelog. Follow the format of the other messages: your change, a link to this PR, and "contributed by <yourname>" with a link to your GitHub profile.

With those changes, it'll be ready to merge 👍

estebanbouza:
Got it, made those changes now, let me know if that works!

itsliamegan:
Looks great!

PR #312, opened 2021-06-08
Fix ProgressReporter accuracy on skipped tests

seven1m:
When `detailed_skip` is turned off (it's on by default), the `ProgressReporter` would not advance the progress bar on skipped tests.


itsliamegan:
Can you post a screenshot/gif of the issue? I can't replicate locally, but looking at the code tells me that this *should* be an issue.

Here's what it looks like with `detailed_skip` enabled:
image

Here's what it looks like with `detailed_skip` disabled:

seven1m:
I'm not at the computer right now, but I believe you only notice it *before* the tests are finished running. That is to say, once the tests are done, the progress bar jumps to 100% no matter what.

In my case, my test suite runs over the course of 5 minutes, and when it gets to about 76%, it immediately jumps to 100% because about 24% of my specs are skips. 😆

itsliamegan:
I was running a small suite, so it makes sense that this wouldn't surface. Good to go then 👍

PR #313, opened 2021-08-12
Add file attribute on JUnit <testcase> tag

nbudin:
CircleCI requires a `file` attribute on the `<testcase>` tag in order to recognize what files each case belong to - it can't use the `filepath` attribute on the `<testsuite>` tag that minitest-reporters puts there already, unfortunately.

They do do this in [their own minitest-ci gem](https://github.com/circleci/minitest-ci/blob/8b72b0f32f154d91b53d63d1d0a8a8fb3b01d726/lib/minitest/ci_plugin.rb#L141), so it would be awesome if minitest-reporters could support it also!


itsliamegan:
Thanks for your contribution! If you can just add a bullet point to the CHANGELOG.md file under the "dev" header—look to the other bullet points for the format—then this is good to merge 👍

nbudin:
> Thanks for your contribution! If you can just add a bullet point to the CHANGELOG.md file under the "dev" header—look to the other bullet points for the format—then this is good to merge 👍

Thanks @itsliamegan!  I've added that to the changelog.

PR #318, opened 2021-12-17
Remove deprecated call to Object#=~

etiennebarrie:
Also updated the Regexp literal to be like minitest's.

### DIFF
--- a/lib/minitest/minitest_reporter_plugin.rb
+++ b/lib/minitest/minitest_reporter_plugin.rb
@@ -41,7 +41,7 @@ module Minitest
       # stolen from minitest self.run
       def total_count(options)
         filter = options[:filter] || '/./'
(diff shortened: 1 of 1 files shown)


etiennebarrie:
Actually that may not be the right fix. We're passing an object as the `:filter` option which does not implement `=~` so it falls back to `Object#=~` which is deprecated. But in the normal case, `:filter` is a String and the current code works well.

Inverting the call results in an error if the Object can't be implicitly converted to a String (does not implement `to_str`).

etiennebarrie:
Ok I found a better fix in minitest: https://github.com/seattlerb/minitest/commit/2a9878be4e332b1d69c6957d9cbb0495a7d7f31f

barrettkingram:
@etiennebarrie thanks! Good idea borrowing the fix from minitest.

PR #324, opened 2022-02-24
Migrate CI to GitHub Actions

petergoldstein:
As Travis CI.org is no longer supported, CI is no longer running in this repo.

This PR switches CI to GitHub Actions.  It maintains all of the Rubies listed in the .travis.yml file except for 1.9.3.  This version was problematic to port, and is long EOLed, so I made a decision to drop it.

Everything runs green on my fork.

PR #325, opened 2022-02-24
Fix RuboCop configuration

andyw8:
The current config is outdated, and fails with:

```
minitest-reporters master % bundle exec rubocop
.rubocop.yml: Metrics/LineLength has the wrong namespace - should be Layout
Error: The `Naming/UncommunicativeMethodParamName` cop has been renamed to `Naming/MethodParameterName`.
(obsolete configuration found in .rubocop.yml, please update it)
`Performance/*` has been extracted to the `rubocop-performance` gem.
(obsolete configuration found in .rubocop.yml, please update it)
`Rails/*` has been extracted to the `rubocop-rails` gem.
(obsolete configuration found in .rubocop.yml, please update it)
```

I've made a number of updates to make it runnable again. Once #324 is merged, this can enabled to run as part of CI.


.rubocop.yml
```diff
@@ -15,22 +19,19 @@ Metrics:
...
   Enabled: true
 
-Naming/UncommunicativeMethodParamName:
+Naming/MethodParameterName:
   MinNameLength: 1
 
 Performance:
   Enabled: true
 
-Rails:
```
andyw8, line 31: No longer need since rubocop-rails is now a separate gem.

os97673:
Do you happen to know when the naming was changed in Rubocop?  What version?

andyw8:
> Do you happen to know when the naming was changed in Rubocop? What version?

I think it should be in the [changelog](https://github.com/rubocop/rubocop/blob/master/CHANGELOG.md).

andyw8:
I see this is failing on some EOL Ruby versions:

https://github.com/andyw8/minitest-reporters/actions/runs/2049705204

What's the policy for supporting these?

os97673:
> I see this is failing on some EOL Ruby versions:
> 
> https://github.com/andyw8/minitest-reporters/actions/runs/2049705204
> 
> What's the policy for supporting these?

Ideally we want to support as many versions as possible.  Do you know what is wrong there?  Is there a way to have logs for actions runs?

andyw8:
I'm seeing an alert "1 workflow awaiting approval" so if you are able to approve that you'll see the full output.

On my branch I'm seeing:

```
      rubocop-performance was resolved to 1.0.0, which depends on
        ruby (>= 2.2.0)
```

But I'm confused as to where that's coming from since there's no dependency on `rubocop-performance` 🤔

andyw8:
Oh, wait, I see...

andyw8:
Ok, I've updated the config and it's passing with all the Ruby versions. I had to remove `rubocop-performance` gem since that doesn't support older Rubies.

os97673:

PR #326, opened 2022-04-08
Fix misspelled method name (`percent_skipps`)

andyw8:


barrettkingram:
@andyw8 Thanks for the contribution 💯 !

Note - I added a method alias for the misspelled version of the method name in case someone is already using it in a custom template with `HtmlReporter`.

PR #331, opened 2022-11-09
add options to reporters for suppressing non-essential output

ashkulz:


ashkulz:
Any chance of reviewing this, @barrettkingram?

barrettkingram:
Can you explain the rationale behind this change? Why is this needed?

ashkulz:
@barrettkingram this PR adds options to remove all test output (via `puts`) -- we run tests in parallel and we'd prefer to only have deprecation warnings show up in the log.

Issue #330, opened 2022-09-21
Support GitHub Action annotations


ekohl:
GitHub supports generating annotations in actions. See https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-notice-message for an example. This would be great. Especially if it's autodetected (using `if ENV['GITHUB_ACTION']` for example. This makes it easier to figure out which tests failed, similar to how xunit can be used in Jenkins.

Earlopain:
One workaround I tried was to use https://github.com/mikepenz/action-junit-report to parse the JUnit report that can be generated. That works somewhat, but not really. The action uses the message property, and that only contains the first line, which means that only the expectation is shown. Not the actual value on the next line.

Rubocop has an integrated actions reporter [here](https://github.com/rubocop/rubocop/blob/2f8239f61bd41fe672a4cd9d6f24e334eba7854a/lib/rubocop/formatter/git_hub_actions_formatter.rb), and there's a [gem](https://github.com/Drieam/rspec-github) for rspec. The code doesn't look complicated, I might take a shot at implementing this

PR #332, opened 2022-11-18
Add Github Actions reporter

Earlopain:
See #330

First commit moves a few functions from invdividual reporters into the base one for reuse. The second commit is the actual reporter. I can drop the first commit if you want that.

Here's a screenshot of how it looks like in Githubs UI:

image

### DIFF
--- a/lib/minitest/reporters/github_actions_reporter.rb
+++ b/lib/minitest/reporters/github_actions_reporter.rb
@@ -1,0 +1,55 @@
+# Inspired by the rubocop formatter.
+# https://github.com/rubocop/rubocop/blob/2f8239f61bd41fe672a4cd9d6f24e334eba7854a/lib/rubocop/formatter/git_hub_actions_formatter.rb
+
+module Minitest
+  module Reporters
+    # Simple reporter designed for Github Actions
+    class GithubActionsReporter < BaseReporter
+      ESCAPE_MAP = { '%' => '%25', "\n" => '%0A', "\r" => '%0D' }.freeze
+
+      def initialize(base_path = Dir.pwd)
+        super({})
+        @base_path = base_path
+      end
+
+      def start
+        super
+
+        print_run_options
+      end
+
+      def record(test)
+        super
+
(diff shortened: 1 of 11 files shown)


jagthedrummer:
This looks great! Would love to see it merged and released.

test/fixtures/github_actions_test.rb
```diff
@@ -0,0 +1,18 @@
+require 'bundler/setup'
+require 'minitest/autorun'
+require 'minitest/reporters'
+require 'minitest/reporters/mean_time_reporter'
```
ekohl, line 4: Why is this mean time reporter required here?

lib/minitest/reporters/github_actions_reporter.rb
```diff
@@ -0,0 +1,50 @@
...
+        end
+      end
+
+      private
+
+      def to_annotation(test, failure_type)
+        message = "#{failure_type}: #{test.name}\n\n#{test.failure.message}"
+        line_number = location(test.failure).split(":").last
+
+        "\n::error file=%<file>s,line=%<line>d::%<message>s" % {
```
ekohl, line 38: In my testing I needed to add a line ending
```suggestion
        "\n::error file=%<file>s,line=%<line>d::%<message>s\n" % {
```

Perhaps `message` below needs to be `message.rstrip` to prevent duplicate line endings as well.

ekohl (commented):
Because I couldn't wait for this to be merged I've created a standalone gem for this: https://github.com/ekohl/minitest_reporters_github / https://rubygems.org/gems/minitest_reporters_github

Once this is merged, I'll retire it.

test/fixtures/github_actions_test.rb
Earlopain: No idea, it's been a hot minute. I have lost interest for this and don't feel like investigating unless there is some feedback from maintainers, at which point I'll be happy to get this in a shape they feel comfortable merging.

seuros:
Can we have an update here ?

PR #342, opened 2023-07-13
Drop support for Ruby < 2.3

barrettkingram:
Ruby 2.2 has been EOL for a long time and the Ruby 2.2 install is broken on GitHub Actions (https://github.com/ruby/setup-ruby/issues/496). Dropping support for Ruby 2.2 and below since we can no longer test against this version reliably.


Nakilon:
So we drop gem compatibilty because the GitHub's Ubuntu can't compile something?...

barrettkingram:
Thanks for the feedback!

My thinking is I don't want to claim this gem is compatible on Ruby 2.2 when we're no longer testing against 2.2. We technically _could_ just stop testing on that version without bumping the required Ruby version. But we would likely introduce an incompatible change eventually, which would be a bad experience for anyone attempting to use this gem on the old version and expecting things to work. Bumping the required Ruby version would allow users of this gem on older versions of Ruby to stick with the latest version of the gem tested for compatibility with their Ruby version.

Plus, there is another reason to drop support for old versions - mainly because it makes it easier on contributors to not have to worry about compatibility going so far back.

I think Windows support is a slightly different issue - I'd prefer to solve that by adding test coverage on a Windows OS. I haven't looked into how that works with GitHub Actions yet but it would be nice to have that some day.

PR #346, opened 2023-08-04
Output failure screenshot when available

matteeyah:
## Summary

Now that `minitest` supports adding custom metadata to tests and their results (https://github.com/minitest/minitest/issues/956) and `rails` automatically adds the failure screenshot path to the metadata (https://github.com/rails/rails/pull/48863) we can surface that metadata as an attachment in the JUnit output.

CI tools use this attachment to display the screenshot when parsing JUnit test reports (e.g. GitLab https://docs.gitlab.com/ee/ci/testing/unit_test_reports.html#view-junit-screenshots-on-gitlab).

This is an example of what it looks like inside of the GitLab UI

Screenshot 2023-08-05 at 15 05 16

### DIFF
--- a/lib/minitest/reporters/junit_reporter.rb
+++ b/lib/minitest/reporters/junit_reporter.rb
@@ -86,31 +86,23 @@ module Minitest
         suite_result = analyze_suite(tests)
         file_path = get_relative_path(tests.first)
 
-        if @timestamp_report
-          xml.testsuite(:name => suite, :filepath => file_path,
-                                    :skipped => suite_result[:skip_count], :failures => suite_result[:fail_count],
-                                    :errors => suite_result[:error_count], :tests => suite_result[:test_count],
-                                    :assertions => suite_result[:assertion_count], :time => suite_result[:time],
-                                    :timestamp => suite_result[:timestamp]) do
-            tests.each do |test|
-              lineno = get_source_location(test).last
-              xml.testcase(:name => test.name, :lineno => lineno, :classname => suite, :assertions => test.assertions,
-                           :time => test.time, :file => file_path) do
-                xml << xml_message_for(test) unless test.passed?
-              end
-            end
-          end
-        else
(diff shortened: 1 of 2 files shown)


matteeyah:
@barrettkingram Could you take a look at this? It's a simple change to the JUnit reporter that adds the screenshot path if it exists?

matteeyah:
@barrettkingram A gentle nudge. Can get this simple change reviewed? It'd add value to a bunch of people, and the implementation is not overly complex 🙃

barrettkingram:
@matteeyah Sorry for the delay and thanks for the contribution! This looks very useful. Nice work coordinating with all the different projects needed to get this working.

One request: I think we should also output the screenshot if the `@timestamp_report` option is set.

Looks like we could refactor that conditional to avoid needing to add multiple calls to `xml_attachment_for`. Maybe something like the following:
```ruby
testsuite_attributes = {
  name: suite,
  filepath: file_path,
  (...)
}
testsuite_attributes[:timestamp] = suite_result[:timestamp] if @timestamp_report
xml.testsuite(testsuite_attributes) do
  tests.each do |test|
    lineno = get_source_location(test).last
    xml.testcase(:name => test.name, :lineno => lineno, :classname => suite, :assertions => test.assertions,
                 :time => test.time, :file => file_path) do
      xml << xml_message_for(test) unless test.passed?
      xml << xml_attachment_for(test) if test.respond_to?('metadata') && test.metadata[:failure_screenshot_path]
    end
  end
end
```

matteeyah:
@barrettkingram Thanks for getting back to me on this. What you suggested makes total sense. I pushed a change with your suggestion.

Side-note: I kept the old formatting style `:key => value` since the rest of the file uses this, and I thought it makes sense to keep consistent about it.

matteeyah:
@barrettkingram Another gentle nudge - hope you don't mind.

PR #347, opened 2023-09-10
Report durations and numbers with consistent significant figures

notEthan:
Currently reporters use a fixed number of digits after a decimal to report test duration and other numbers, though the number of digits varies from 0 (whole numbers) all the way to 6. Since tests have a huge range of execution time, this can result in not-very-useful numbers in both directions: giving microsecond precision on a test that takes two minutes is unnecessary information that's harder to parse; on the other hand a test that takes some microseconds is, in some places, truncated to 0.00.

This PR changes reporting to include 3 [significant figures](https://en.wikipedia.org/wiki/Significant_figures) on all numbers (barring any that I missed). 3 seems like at least as much precision as is significant in these reports - the amount of time tests take always varies more than this (in my experience).

I also split durations into hour/minute/second rather than always just reporting seconds.

Before:

```
test_0001_fast_test   PASS (0.00s)
test_0002_slow_test   PASS (105.00s)

Finished in 105.09427s
```

After:

```
test_0001_fast_test   PASS (0.00114s)
test_0002_slow_test   PASS (1m 45s)

Finished in 1m 45s
```

MeanTimeReporter before:

```
Finished tests in 103.051912s, 0.0194 tests/s, 1.0092 assertions/s.
2 tests, 102 assertions, 0 failures, 0 errors, 0 skips

Minitest Reporters: Mean Time Report (Samples: 5, Order: :avg :desc)
Avg: 102.81607025 Min: 101.051376   Max: 105.07253    Last: 103.051096   Description: slow
Avg: 0.001024625  Min: 0.000411     Max: 0.001939     Last: 0.000629     Description: fast
```

After:

```
Finished tests in 1m 44s, 0.0192 tests/s, 1.01 assertions/s.
2 tests, 102 assertions, 0 failures, 0 errors, 0 skips

Minitest Reporters: Mean Time Report (Samples: 5, Order: :avg :desc)
Avg: 102       Min: 101        Max: 105       Last: 104        Description: slow
Avg: 0.00115   Min: 0.000411   Max: 0.00194   Last: 0.000437   Description: fast
```

### DIFF
--- a/lib/minitest/reporters/base_reporter.rb
+++ b/lib/minitest/reporters/base_reporter.rb
@@ -61,6 +61,15 @@ module Minitest
         end
       end
 
+      def print_finished
+        puts('Finished in %.5fs' % total_time)
+        print('%d tests, %d assertions, ' % [count, assertions])
+        color = failures.zero? && errors.zero? ? :green : :red
+        print(send(color, '%d failures, %d errors, ') % [failures, errors])
+        print(yellow '%d skips' % skips)
+        puts
+      end
+
       protected
 
       def after_suite(test); end
--- a/lib/minitest/reporters/base_reporter.rb
+++ b/lib/minitest/reporters/base_reporter.rb
@@ -62,7 +62,7 @@ module Minitest
       end
 
       def print_finished
-        puts('Finished in %.5fs' % total_time)
+        puts('Finished in %s' % format_duration(total_time))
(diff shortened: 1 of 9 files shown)


notEthan:
I notice this would conflict with https://github.com/minitest-reporters/minitest-reporters/pull/340 but it could supersede that. This does not have the round issue that PR fixes.

PR #348, opened 2023-09-10
ProgressReporter fix backwards fraction in format

notEthan:
This was "total/completed" instead of "completed/total"


ktreis:
See #350 for some additional history/justification for this change

barrettkingram:
Thank you @notEthan and @ktreis !

PR #306, opened 2020-12-03
Print errors at the bottom when using Spec reporter

picandocodigo:
This is an initial attempt for a modification to the Spec reporter that is much more convenient for me personally. I'm using this reporter in [`elasticsearch-xpack`](https://github.com/elastic/elasticsearch-ruby/tree/master/elasticsearch-xpack), and it would be much more practical for me to see the errors at the bottom. With this change, the error is still reported with the rest of the tests:
image

But you can see the detail in errors or failures at the bottom of the report:
image

Would this be convenient to add? Maybe as an option?

Related: #304

PR #352, opened 2024-05-23
Adds explicit option to `suppress_inline_failure_output` to `SpecReporter`

rthbound:
Provides an explicit switch, separate from `print_failure_summary`, for suppressing inline failure messages.

  - Until this change, the `print_failure_summary` would do both: print a failure/error summary after all tests run and suppress the inline failure messages.
  - With this change `print_failure_summary` will just add a summary at the end of a test run, and it will no longer suppress the inline failure messages.

  The summary added in #306 is indeed helpful to prevent need for scanning a large number of
  tests' output.

  The inline output is helpful for immediately getting to work on
  addressing some test failure without waiting for all tests to finish.

  Our project opted in to `print_failure_summary` (which is nice) but
  I found the inconvenience of needing to wait for the tests to
  conclude before taking action to be a great hindrance.


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
...
 * Updated `JUnitReporter` to output a failure screenshot path when included in the minitest result [#346](https://github.com/minitest-reporters/minitest-reporters/pull/346) contributed by [matteeyah](https://github.com/matteeyah)
 * Fixed backwards fraction in `ProgressReporter` default format [#348](https://github.com/minitest-reporters/minitest-reporters/pull/348) contributed by [notEthan](https://github.com/notEthan)
+* Added option `suppress_inline_failure_output` to `SpecReporter` to provide an explicit switch, separate from `print_failure_summary`, for suppressing inline failure messages. Until this change, the `print_failure_summary` would do both: print a failure/error summary after all tests run and suppress the inline failure messages. With this change `print_failure_summary` will just add a summary at the end of a test run, and it will no longer suppress the inline failure messages.
+  as they happen [#306](https://github.com/minitest-reporters/minitest-reporters/pull/306) contributed by [picandocodigo](https://github.com/picandocodigo)
```
barrettkingram, line 9:
```suggestion
```

barrettkingram:
I think this is a good improvement! Can you update the doc annotations for `SpecReporter#initialize` so the new option is documented and the modified behavior of `print_failure_summary` is accurate?

PR #353, opened 2024-06-07
feat: Add GithubReporter to Minitest::Reporters

kylewelsby:
This commit adds a new reporter, GithubReporter, to the Minitest::Reporters module. The GithubReporter is responsible for recording test results and outputting them in a format compatible with GitHub Actions. It determines the type of the test (notice, error, or skipped) and creates an output string containing the file, line, and title of the test. The output is then printed to the console.

The GithubReporter is implemented as a subclass of BaseReporter and includes methods for determining the type of the test, creating the output string, and escaping special characters in the output. It also includes a private method for converting absolute file paths to relative paths.

This commit also includes the necessary changes to the lib/minitest/reporters.rb file to autoload the GithubReporter class.

The addition of the GithubReporter provides a new way to report test results in a format that is compatible with [GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message), making it easier to integrate Minitest with CI/CD pipelines.

### DIFF
--- a/lib/minitest/reporters/github_reporter.rb
+++ b/lib/minitest/reporters/github_reporter.rb
@@ -1,0 +1,45 @@
+module Minitest
+  module Reporters
+    class GithubReporter < BaseReporter
+      def record(test)
+        super
+
+        type = determine_type(test)
+        output = create_output(test)
+        message = test.failure
+
+        puts "#{type} #{output.join(',')}::#{message}"
+      end
+
+      private
+
+      def determine_type(test)
+        if test.skipped?
+          "::notice"
+        elsif test.error? || test.failure
+          "::error"
+        end
+      end
+
+      def create_output(test)
+        {
+          file: relative_path(test.source_location[0]),
+          line: test.source_location[1],
+          title: test.name,
+        }.map { |k, v| "#{k}=#{escape_properties(v)}" }
+      end
+
+      def relative_path(path)
+        Pathname.new(path).relative_path_from(Pathname.new(Dir.getwd))
+      end
+
+      def escape_properties(string)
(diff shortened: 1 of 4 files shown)


lib/minitest/reporters/github_reporter.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      def escape_properties(string)
+        string.to_s.gsub("%", '%25')
+              .gsub("\r", '%0D')
+              .gsub("\n", '%0A')
+              .gsub(":", '%3A')
+              .gsub(",", '%2C')
+      end
+
+      def location(exception)
```
erickguan, line 53: Maybe use [`Exception#backtrace_locations`](https://www.rubydoc.info/stdlib/core/Exception:backtrace_locations) or  `Exception#backtrace`.

erickguan (commented):
Lovely PR! I really want to see this PR getting merge soon.

lib/minitest/reporters/github_reporter.rb
```diff
@@ -0,0 +1,66 @@
...
+        last_before_assertion = ''
+        exception.backtrace.reverse_each do |s|
+          break if s =~ /in .(assert|refute|flunk|pass|fail|raise|must|wont)/
```
evgeni: Ruby 3.4 includes the owner in the backtrace (see https://bugs.ruby-lang.org/issues/19117)
You can catch both styles with
```suggestion
          break if s =~ /in .(Minitest::Assertions#)?(assert|refute|flunk|pass|fail|raise|must|wont)/
```

PR #354, opened 2024-06-20
Fix logic of new suppress_inline_failure_output option

barrettkingram:
`suppress_inline_failure_output` was actually doing the opposite of what it claimed to do. This commit reverses that.


rthbound:
🤕 Sorry about that. Thanks for seeing it through.

PR #357, opened 2024-10-13
use module prepend instead of alias method chaining

grosser:
- if another library uses prepend on `run` we get into an endless loop
- prepend is available in all non-EOL ruby versions
- prepend is cleaner
needs https://github.com/minitest-reporters/minitest-reporters/pull/358

### DIFF
--- a/lib/minitest/reporters.rb
+++ b/lib/minitest/reporters.rb
@@ -41,21 +41,18 @@ module Minitest
     end
 
     def self.use_around_test_hooks!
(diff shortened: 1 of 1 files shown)


connorshea:
Thank you for this fix! It fixes a problem with running this gem and minitest-retry together 🙇

mikevoets:
Thank you! @barrettkingram do you think this will be released sometime soon? No pressure 🙏🏻

PR #358, opened 2024-10-13
drop legacy ruby versions

grosser:
even if there is anyone running 2.2, they do not need the latest minitest-reporters
reduce maintenance toil from failing PR and allow using modern ruby syntax [example](https://github.com/minitest-reporters/minitest-reporters/pull/357) `Module.prepend`


seuros (approved):

grosser:
@barrettkingram can you check out this PR + restart the failing built plz :)

andyw8:
Dropping older versions would also help with https://github.com/minitest-reporters/minitest-reporters/pull/362

.rubocop.yml
```diff
@@ -1,6 +1,7 @@
 ---
 AllCops:
   NewCops: enable
+  TargetRubyVersion: 2.3 # keep in sync with gemspec and ci.yml
```
barrettkingram: It seems Rubocop [should detect this automatically](https://docs.rubocop.org/rubocop/latest/configuration.html#setting-the-target-ruby-version). So I lean towards avoiding specifying it in this file - one less thing to have to keep in sync when the version updates in the future.

barrettkingram:
@grosser Thanks for the contribution! I agree it makes sense to drop those old ruby versions, especially in 2026

